PR #8, opened 2020-03-19
Derive Session Keys

lthiery:
This provide extra parser functions that help a user derive session keys. I've tested this on live packets and it works.

I can add tests tomorrow if you like.

### DIFF
--- a/src/creator.rs
+++ b/src/creator.rs
@@ -412,7 +412,7 @@ impl DataPayloadCreator {
     ///     .set_channel_mask_ack(true)
     ///     .set_data_rate_ack(false)
     ///     .set_tx_power_ack(true);
-    /// let cmds: Vec<&lorawan::maccommands::SerializableMacCommand> = vec![&mac_cmd1, &mac_cmd2];
+    /// let cmds: Vec<&dyn lorawan::maccommands::SerializableMacCommand> = vec![&mac_cmd1, &mac_cmd2];
     /// phy.set_mac_commands(cmds);
     /// ```
     pub fn set_mac_commands<'a>(
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -127,6 +127,10 @@ impl <'a, T: AsRef<[u8]>> GenericPhyPayload<T> {
         Ok(result)
     }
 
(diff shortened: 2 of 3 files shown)


lthiery:
I'm also sneaking in a helper that allows me to access the raw data inside the GenericPayload. I'm using it do so something like this:

```
match packet.mac_payload() {
    MacPayload::JoinAccept(_) => {
        let key = [
            0xEB, 0x31, 0xA2, 0x94, 0x00, 0x51, 0x7C, 0x53, 0x12, 0xCF, 0xBF, 0xD5,
            0xF5, 0x6F, 0x69, 0xC2,
        ];
        let app_key = keys::AES128(key);
        let decrypted_join_accept =
            GenericPhyPayload::<[u8; 17]>::new_decrypted_join_accept(
                packet.inner_ref().clone(),
                &app_key,
            )
            .unwrap();
```

ivajloip:
Hello Louis,

Thank you for this suggestion - now that you mention it, I agree this could be really useful addition :slightly_smiling_face: . If you would have the time to add some tests, that would be great. Otherwise I will try to spend some time on that tomorrow. 

Another point that is related to the issue you opened ( #9 ) - where should this functionality be placed? One option is inside the decrypted join accept. Another one is to remain functions - maybe in the same module or not. Currently it would seem to me that having it inside a join accept might be better, but I would be curious to know your take on that.

Thanks,
Ivaylo

lthiery:
Since this takes information from JoinRequest and JoinAccept, I would keep it as is more or less.

One thing that could be different with the types would be to simplify the signature a little bit by passing the MTypes:

```
derive_appskey(join_request: &JoinRequest, 
    join_accept: &DecryptedJoinAccept,
    key: &keys::AES128) 
```

When I see this use case, I lean towards doing MTypes variants that are Decrypted vs Encrypted (rather than making their payloads be of type `Encrypted([u8])` or `Decrypted([u8])`) as mentioned in https://github.com/ivajloip/rust-lorawan/issues/9.

lthiery:
By the way, I've added the tests :)

ivajloip:
Thanks :slightly_smiling_face:

PR #18, opened 2020-05-08
feature gate use JoinAccept

lthiery:


src/default_crypto.rs
```diff
@@ -10,7 +10,11 @@ use aes::{Aes128, block_cipher_trait::BlockCipher};
 use generic_array::{GenericArray, typenum::U16};
 
 use super::keys::*;
-use super::creator::{JoinAcceptCreator, JoinRequestCreator, DataPayloadCreator};
+use super::creator::{JoinRequestCreator, DataPayloadCreator};
+
+#[cfg(feature = "with-uplink")]
```
ivajloip, line 15: Shouldn't this be `with-downlink`

lthiery: of course it should! heh

ivajloip:
Thanks :)

PR #22, opened 2020-06-03
Give device stack its own buffer to simplify Radio trait

lthiery:
The song and dance around getting buffers between the Radio and the device stack seemed a little much just to save a little bit of RAM.


lthiery:
I'm working on a larger refactor that will make this irrelevant

PR #23, opened 2020-06-09
Fix Mac Payloads

lthiery:
I came across an interesting bug when I was debugging a device's LinkADRReq process with a server. [The specification](https://lora-alliance.org/sites/default/files/2018-05/lorawan1_0_2-20161012_1398_1.pdf) for LinkADRReq is kind of interesting (p.24 l. 33-36):

> The network server may include multiple LinkAdrReq commands within a single downlink message. For the purpose of configuring the end-device channel mask, the end-device will process all contiguous LinkAdrReq messages, in the order present in the downlink message, as a single atomic block command. 

When I listed the FOpts from the downlink, I was seeing:
```
LinkADRReq(LinkADRReqPayload([0, 0, 0, 112, 3, 0, 0, 255, 0]))
LinkADRReq(LinkADRReqPayload([0, 0, 255, 0]))	
```

While multiple LinkADRReq payloads were being parsed, the payload includes the second payload, unfortunately.

The small correction to the `parse_one_mac_cmd` macro resolves the issue. I've added a test case to cover the case as well.

### DIFF
--- a/encoding/src/maccommands.rs
+++ b/encoding/src/maccommands.rs
@@ -198,7 +198,7 @@ macro_rules! mac_cmds {
         fn parse_one_mac_cmd<'a, 'b>(data: &'a [u8], uplink: bool) -> Result<(usize, MacCommand<'a>), &'b str> {
             match (data[0], uplink) {
                 $(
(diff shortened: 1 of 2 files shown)


ivajloip:
Looks good, thanks for catching this :slightly_smiling_face:

PR #24, opened 2020-06-25
Update Cmac

lthiery:


encoding/Cargo.toml
```diff
@@ -9,9 +9,8 @@ license = "MIT"
 keywords = ["lorawan", "iot", "lpwan", "parser", "lightweight"]
 
 [dependencies]
-aes = { version = "0.3.2", optional = true }
-cmac = { version = "0.2.0", optional = true }
-generic-array = "0.12.0"
+aes = { version = "0.4.0", optional = true }
+cmac = { version = "0.3.0", optional = true }
```
ivajloip, line 13: I am not sure how `aes` could be an optional dependency in case `encoding/src/securityhelpers.rs` relies on it. My intention is to keep `aes` as truly optional dependency and that is why generic-array should be kept in my opinion.

lthiery: Fair points! Forgot about that when I was making this PR. Will fix tomorrow hopefully

lthiery: Alright, I think it's been done. Let me know if you spot anything else

lthiery: Shoot - it looks like some formatting fixes got passed along too. I think this is needed to pass CI though.

Also, just a note, I added another test case for CI to test for --no-default features. That should keep the mistake I made from happening again.

ivajloip (approved):
Looks good to me :slightly_smiling_face: Thank you!

ivajloip:
That change has been merged as part of #28 , therefore closing. Thanks again!

PR #28, opened 2020-07-22
Device Stack Refactor

lthiery:
The design of the device stack is more or less entirely overhauled in favor of a more structured and "Rustic" state machine design. The design and state of the stack is more fully discussed in the `device/README.md`.

In addition, the embedded concerns have been removed from this project in favor of pushing them into [sx12xx-rs](https://github.com/lthiery/sx12xx-rs). This will make this allows this repository to be little more streamlined.


device/src/lib.rs
```diff
@@ -1,334 +1,234 @@
...
-                    radio.send_buffer();
-                    self.sm_handler = Device::joined_sending;
-
-                    None
+    pub fn get_downlink_payload(&mut self) -> Option<(u8, Vec<u8, U256>)> {
+        let buffer = self.get_radio().get_received_packet();
+        if let Ok(parsed_packet) = lorawan_parse(buffer, C::default()) {
+            if let PhyPayload::Data(data_frame) = parsed_packet {
+                let fport = data_frame.f_port();
+                if let DataPayload::Decrypted(decrypted) = data_frame {
```
ivajloip, line 214: I am surely missing something, but when will the `data_frame` be already decrypted? I don't think `lorawan_parse`/ `parse_with_factory` can produce a decrypted payload as it does not know the keys.

ivajloip (commented):
Thank you for this rework! It seems very nice to me. I checked is some details everything except the state machines changes, where I checked very little. I will try to finish that in the following few days. In the meantime, please check my comment as either I am not understanding something, or there might be something in the code that is missing :slightly_smiling_face:

device/src/lib.rs
lthiery: Good catch! This is actually something I only noticed Friday that it was being parsed as an encrypted payload here despite the payload having been decrypted in the state machine. I'm working on a fix for this soon.

My idea is to have an enum for the send/receive buffer and to store the decrypted structure there after decryption.

lthiery: I had trouble using the same memory via enum so I just made a copy for now. I'll consider making it an enum to be an optimization for later :P

device/src/state_machines/session.rs
```diff
@@ -0,0 +1,632 @@
...
+                                                    )
+                                                    .unwrap();
+
+                                                let mut copy = Vec::new();
+                                                copy.extend(decrypted.as_bytes());
+
+                                                // this is a sane unwrap because we know the decrypted
+                                                // payload is a valid DecryptedDataPayload
+                                                self.shared.data_downlink = Some(
+                                                    DecryptedDataPayload::new_from_decrypted_payload(
```
ivajloip, line 504: My understanding is that the copy is created here in order to be able to reuse the `data_frame` bytes for other purposes and in order to be able to keep the downlink in a field. If that is correct, I am wondering what is the advantage of this implementation over one where the copying is done earlier and the parsing logic has bytes that are owned right from the start. My reasoning is that I am not too fond of the idea of exposing an interface to create `DecryptedDataPayload` from raw bytes. I believe it could cause some confusion and lead to some errors if people are not careful when they provide encrypted data and when unencrypted one. That is to a great extend the reason to have two separate types - encrypted and unencrypted one. For that reason, I would rather not have such a method available, unless it is crucial for some use case. Other than those changes, I don't see any blocking points.

lthiery: I think that's a great idea. Thanks!

ivajloip:
Thank you for your contribution and for working with me to have it well aligned with my vision of the encoding API :slightly_smiling_face:

PR #29, opened 2020-09-30
Fix FCntUp increment logic

lthiery:
This fixes an issue where FCnt was automatically incremented. In the case of a confirmed frame that does not receive an ACK, the FCnt should not be incremented, as per the specification:
```
The FCnt is not incremented in case of multiple transmissions of an unconfirmed frame (see NbTrans parameter), or in the case of a confirmed frame that is not acknowledged.
```

The re-transmissions is left to the user, ie: they need to send the same payload to implement a re-transmit after the NoAck response from the stack. This at least makes this possible as the FCnt is otherwise inaccessible to the user.

### DIFF
--- a/device/src/state_machines/session.rs
+++ b/device/src/state_machines/session.rs
@@ -197,8 +197,6 @@ where
             }
         }
 
-        self.session.fcnt_up_increment();
-
         match phy.build(
             &data.data,
             dyn_cmds.as_slice(),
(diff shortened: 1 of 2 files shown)


device/src/state_machines/session.rs
```diff
@@ -485,6 +483,12 @@ where
                                                 && (fcnt > session.fcnt_down || fcnt == 0)
                                             {
                                                 session.fcnt_down = fcnt;
+                                                // increment the FcntUp since we have received
+                                                // downlink - only reason to not increment
+                                                // is if confirmed frame is sent and no
+                                                // confirmation (ie: downlink) occurs
+                                                session.fcnt_up_increment();
```
ivajloip, line 490: am I right to think that here is an assumption that the device is a class A device and if so is it an intended assumption and shouldn't it be documented?

lthiery: Yes, that's absolutely an assumption! I'll add it to the docs - good point

ivajloip:
Thanks!

PR #31, opened 2020-11-21
Crypto Fix

lthiery:
This relates to [the issue filed](https://github.com/ivajloip/rust-lorawan/issues/30) this morning. I noticed that the function `encrypt_frm_data_payload` looks more or less modelled after [PayloadEncrypt from the Loramac Node repo](https://github.com/Lora-net/LoRaMac-node/blob/master/src/mac/LoRaMacCrypto.c#L232). I tried to make the code reflect it more 1-to-1 just for the sake of finding the issue.

While this "fixes" it right now, I think I want to rework this slightly more. I think the core issue lies in the fact that the helper_block is being modified in place during the encryption (`aes_enc.encrypt_block`) while in Loramac Node, the helper block is used repeatedly to derive a new encrypted block. Therefore, I think changing `aes_enc.encrypt_block` to take an input and output block would be better than generating a new helper block every time around this loop.

Anyway, I'll try to put more work into this in the coming days but I wanted to share the progress so we don't duplicate effort @ivajloip

### DIFF
--- a/device/src/state_machines/session.rs
+++ b/device/src/state_machines/session.rs
@@ -489,7 +489,6 @@ where
                                                 // confirmation (ie: downlink) occurs
                                                 session.fcnt_up_increment();
 
-
                                                 let mut copy = Vec::new();
                                                 copy.extend(encrypted_data.as_bytes());
 
@@ -498,7 +497,10 @@ where
                                                 //      always work when copy bytes from another EncryptedPayload
                                                 // * the decrypt will always work when we have verified MIC previously
                                                 let decrypted =
-                                                    EncryptedDataPayload::new_with_factory(copy, C::default())
+                                                    EncryptedDataPayload::new_with_factory(
+                                                        copy,
+                                                        C::default(),
+                                                    )
                                                     .unwrap()
                                                     .decrypt(
                                                         Some(&session.newskey()),
@@ -571,7 +573,6 @@ where
                     }
                     // Timeout during second RxWindow leads to giving up
                     RxWindow::_2(_) => {
-
                         if !self.confirmed {
                             // if this was not a confirmed frame, we can still
                             // increment the FCnt Up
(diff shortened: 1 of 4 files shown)


lthiery:
@ivajloip i think this is the minimum amount of changes needed to fix the encryption operation. It was actually two issues:

1) The helper block needs to not get encrypted so it can continue being used; I make a copy of it before the encryption operation.
2) The last byte of the encrypted block can't be the index of the entire payload but instead need to reflect the block number

Concerning 1, ideally the crypto trait would allow for an immutable input block and a mutable output block. This would allow crypto functions that take an immutable input (`a_block`) and outputs the encrypted block in an output buffer (`s_block`). This is how the [Loramac Node implementation has it](https://github.com/lthiery/rust-lorawan-crypto/blob/master/lorawan-crypto-sys/lorawan-crypto/aes.c#L567).

But that's a bigger change and I'm not sure if its worth it, so I'll leave this PR with the minimum changes.

device/src/state_machines/session.rs
```diff
@@ -580,7 +581,7 @@ where
 
                         let response = if self.confirmed {
                             Ok(Response::NoAck)
-                            // check if FCnt is used up
+                        // check if FCnt is used up
```
ivajloip, line 584: That seems misindented

ivajloip (commented):
Thank you for looking into this! You have likely found an issue, so thank you! There is a question about adding extra tests, but otherwise it looks good to me.

ivajloip:
Hi,

Sorry for the delay! I addressed the 1-2 points and added the tests, everything seems to be working well now. Because of the tests I decided to merge it locally and push it, which did not close the PR, so closing manually.

Thanks again for reporting this and providing the fix :smile:

lthiery:
Great news! And sorry I never got around to it. Meant to get to it over the holiday break but glad you did!

ivajloip:
No worries, I also meant to get this done and release the new versions earlier, but I had a few other more pressing things to finish. Thanks again for all your help and happy new year :)

PR #32, opened 2021-03-11
Add support for multiple regions as compile time features

lulf:
* Add features to select the an instance of region
* Add frequency plan for EU868 region
* Make more radio properties configured from region

With this I am able to use rust-lorawan with the things network in EU frequency bands. There are a few things like RX2 downstream for EU that I don't think will work with this code, but I'm still a LoRa novice so I'm not sure if its acceptable.

Also, I've tried implementing this both using enums, traits and crate features, and the reason for ending up on using features is that I don't think you can reconfigure the region settings anyway, so it seemed like best approach at present.

### DIFF
--- a/device/Cargo.toml
+++ b/device/Cargo.toml
@@ -17,3 +17,9 @@ lorawan-encoding = { path = "../encoding", default-features = false }
 heapless = "0.5.4"
 as-slice = "*"
 generic-array = "0.14.2"
+
+[features]
+default = ["region+us915"]
+
+"region+us915" = []
+"region+eu868" = []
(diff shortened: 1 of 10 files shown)


ivajloip:
Thank you for your contribution! I imagine for now it is completely fine to have this as a compile time feature. I would prefer in the longer term to be able to dynamically switch between regional parameters, but for the time being the implementation appears perfectly good :slightly_smiling_face:

lthiery:
Agree that it would be great to have this be dynamic. I have some work on this somewhere as well, but I haven't had time to clean it up.

https://github.com/helium/rust-lorawan/tree/lthiery/configurable-regions

PR #33, opened 2021-03-18
Make Region Runtime Configurable

lthiery:
I think we agree that we want selectable regions, so I wanted to remove the feature-flags introduced by #32 

This stuff isn't fully tested yet, but I think it provides a more scalable approach to region handling.

It also implements the European spec more fully by using CF list rather than hardcoded channels.


lthiery:
Reviewing these diffs, I see that I missed a lot of good stuff from @lulf  regarding bandwidth, dbm, and coding rate. I will pull those in

lthiery:
Okay, I think this is in a decent state now and is ready for review.

Basically, I implemented a RegionHandler trait which each region should support. A struct "Configuration" holds the currently configured region and dispatches appropriately based on state. This type of design was favored to avoid dynamic dispatch which has always caused me problems in embedded environments. 

Currently, you can't actually dynamically changed state (I want to make sure its only allowed in certain conditions conditions, ie: not mid-transmit), but the way the code is setup, adding the feature becomes straightforward.

In addition to making region runtime selectable, I did a few more things:
* allow user to take the join accept for inspection/debug
* made EU regional support use CFList instead of hardcoded channel plan
* refactored default network constants to reduce repetition, using default trait
* added some semblance of CN470 support (need a lot more testing)

device/src/state_machines/no_session.rs
```diff
@@ -323,9 +314,9 @@ where
             Event::TimeoutFired => {
                 let rx_config = radio::RfConfig {
                     frequency: self.shared.region.get_join_accept_frequency1(),
-                    bandwidth: self.shared.region.get_bandwidth(),
-                    spreading_factor: self.shared.region.get_spreading_factor(),
-                    coding_rate: self.shared.region.get_coding_rate(),
+                    bandwidth: radio::Bandwidth::_500KHZ,
```
lulf, line 317: These should use the settings from the region I think?

device/src/state_machines/session.rs
```diff
@@ -383,8 +372,8 @@ where
             Event::TimeoutFired => {
                 let rx_config = radio::RfConfig {
                     frequency: self.shared.region.get_rxwindow1_frequency(),
-                    bandwidth: radio::Bandwidth::_125KHZ,
-                    spreading_factor: radio::SpreadingFactor::_7,
+                    bandwidth: radio::Bandwidth::_500KHZ,
```
lulf, line 375: Use settings from region?

lulf (commented):
Looks good to me at least, there are a few places where it should use settings from config instead of hardcoded values (I _think_).

device/src/state_machines/no_session.rs
lthiery: > Looks good to me at least, there are a few places where it should use settings from config instead of hardcoded values (I _think_).

Thanks looks like I missed a few of these!

lthiery:
@lulf thanks for the catches. I also realized that we're doing commit battles on the "Default" implementation for RfConfig/RxConfig so I removed them entirely as they were no longer useful anyway.

lthiery:
After some further testing, there's something wrong with the US version atm. Do not merge, please.

ivajloip:
OK, I will wait for you to let me know when you have found the issue. And thank you both for your collaboration and contribution!

device/src/region/eu868/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+                cf_list[channel - JOIN_CHANNELS.len()]
+            }
+        } else {
+            let channel = random as usize % JOIN_CHANNELS.len();
+            JOIN_CHANNELS[channel]
+        }
+    }
+
+    fn get_join_accept_frequency1(&self) -> u32 {
+        JOIN_CHANNELS[self.last_tx.1 as usize]
```
ivajloip, line 67: I did not notice where `last_tx` is set in this type. Did I miss it? :slightly_smiling_face: Could you explain what is the reason to have `last_tx` be part of the `RegionHandler`? Intuitively it feels better for me to be part of the device state somehow.

device/src/region/eu868/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+        &mut self,
+        join_accept: &super::DecryptedJoinAcceptPayload<T, C>,
+    ) -> JoinAccept {
+        let mut new_cf_list = [0, 0, 0, 0, 0];
+        if let Some(cf_list) = join_accept.c_f_list() {
+            for (index, freq) in cf_list.iter().enumerate() {
+                new_cf_list[index] = freq.value();
+            }
+        }
+        self.cf_list = Some(new_cf_list);
```
ivajloip, line 33: What is the advantage of having cf_list in the region parameters? My first thought would have been that the device copies this in its state while starting and never looks back to the region. Then such changes would happen on the device state.

ivajloip (commented):
Thank you for this work! I have a few questions in order to better understand the change.

device/src/region/cn470/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn set_subband(&mut self, _subband: u8) {}
+
+    fn get_join_frequency(&mut self, random: u8) -> u32 {
+        let channel = random % 96;
+        self.last_tx = channel;
+        UPLINK_MAP[channel as usize]
+    }
+
+    fn get_data_frequency(&mut self, random: u8) -> u32 {
+        let channel = random & 0b111;
```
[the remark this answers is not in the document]
lthiery: That was a mistake it seems. It is now:
```
let channel = random as usize % UPLINK_MAP.len();
```

device/src/state_machines/mod.rs
```diff
@@ -2,18 +2,27 @@ use super::*;
...
+}
+
+enum Downlink {
+    Data(DecryptedDataPayload<Vec<u8, U256>>),
+    Join(JoinAccept),
+}
+
+#[derive(Debug)]
+pub struct JoinAccept {
+    pub cflist: Option<[u32; 5]>,
```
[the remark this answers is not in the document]
lthiery: I'm trying to provide visibility to the user for debug. I think I tried using the type from `lorawan_encoding` but it felt burdensome because of the crypto.

device/src/region/eu868/mod.rs
lthiery: Nope, I missed it! I put it in.

In general, I'm trying to keep anything that's considered "regional lorawan spec" inside of the region module, and anything that's considered part of the core spec in the device state machine. IMO, that should includes state as well as logic.

device/src/region/eu868/mod.rs
lthiery: As I mentioned before, I consider this to be part of "regional logic" as opposed to "core logic". 

Basically, anything where the idea comes from [the regional spec](https://lora-alliance.org/wp-content/uploads/2020/11/lorawan_regional_parameters_v1.0.3reva_0.pdf) will go in regions and anything that's part of the [core spec](https://lora-alliance.org/resource_hub/lorawan-specification-v1-0-3/) goes into the device session/nosession state machines.

lthiery:
The bug had to do with my bandwidth being mis-configured for the downlink in the US. I took the opportunity to try to implement some ideas around Datarate and to start cleaning up the interfaces for RegionHandler.

@lulf can you confirm that it works in EU?

I've updated my example project if that helps: https://github.com/lthiery/rust-sx12xx/tree/lthiery/updates-again

lulf:
@lthiery RX from join seems to need a lot more retries, and downlink data is not working :/

lthiery:
It's very strange that its somehow more flaky. I would think the refactors I did either make it work or not work. I don't think I've impacted the timing at all... but maybe I need to think more on it.

Perhaps I've messed up your spreading factor or bandwidth selection for your region compared to what we were doing before? I defaulted the DR to 0, which for Europe is: `Datarate { spreading_factor: SpreadingFactor::_12, bandwidth:  Bandwidth::_125KHZ }`. It seems that before, you were running: 
```
            bandwidth: Bandwidth::_125KHZ,
            spreading_factor: SpreadingFactor::_7,
```
I've seen some devices with oscillator issues struggle running at SF12, so perhaps that is the issue? I'll add a function so that the user can change the datarate.

What environment are you running this in, out of curiosity? Looks different than the RTIC app in the example. 

It's not terrible well documented at this point, but there's a Timing trait that is coupled with the Radio implementation in `rust-sx12xx`. It defines: 
- how many seconds before the RX window the stack should wake itself up (with a TimerRequest)
- how many seconds after starting the RX window should close itself

It's currently configured as follows (but I should really make it easy to configure)
```
impl lorawan_device::Timings for LorawanRadio {
    fn get_rx_window_offset_ms(&self) -> i32 {
        -500
    }
    fn get_rx_window_duration_ms(&self) -> u32 {
        800
    }
}
```

Therefore, assuming the device is good at marking the end of the transmit (I do it by setting the timer when catching the DIO0 interrupt), there's a generous 500 ms before the window and 300 ms after the window where the radio _could_ be listening. That being said, the 500 ms before the window need to account for dispatching the setup commands and the radio settling into the appropriate configuration.

lthiery:
@lulf not sure if you're notification settings let you know about my request above. Basically, I think you should `device.set_datarate(5)` and see if performance reverts back to what you were used to.

lulf:
@lthiery Sorry for not responding. You're right, I used SF7 earlier, so I will try setting datarate and report back (I'll be able to test this out first thing next week). My application uses the same window settings as your example, so that part is fine I think (at least I did not change this switching from master branch to this branch).

I'm using https://github.com/drogue-iot/drogue-device/ (I'm working on a lora driver that uses this crate + a sx1276 radio, and an example for the stm32 discovery board)

lthiery:
> @lthiery Sorry for not responding. You're right, I used SF7 earlier, so I will try setting datarate and report back (I'll be able to test this out first thing next week). My application uses the same window settings as your example, so that part is fine I think (at least I did not change this switching from master branch to this branch).
> 
> I'm using https://github.com/drogue-iot/drogue-device/ (I'm working on a lora driver that uses this crate + a sx1276 radio, and an example for the stm32 discovery board)

A sx127x driver natively in Rust! Nice.

lulf:
> @lulf not sure if you're notification settings let you know about my request above. Basically, I think you should `device.set_datarate(5)` and see if performance reverts back to what you were used to.

I can confirm that this resolves the issue and RX is working again. Also, I had to make the method pub to be able to use it. 

From my interpretation of the spec, there is no default SF for EU, so I'm fine with whatever value, though perhaps it should be set to what is most likely to work for most radios?

lthiery:
@lulf great news! Insofar as the spec, I recall somewhere it being said "the device SHALL send a Join Request on DR0" so I defaulted to 0. I admit, I'm having trouble finding the reference right now.

In theory, I think low datarates are considered most conservative since spreading factor is highest. Of course, this causes issues when either lots of collisions are possible or when oscillators have trouble maintaining a precise timing over such a long transmit... 

Anyway, I'm open to other approaches long-term. Insofar as this PR, I think leaving a setter accessible can allow for customization by the user.

Also, FYI, I'm working on a quick README update which, amongst other things, will reference your project if that's OK?

lulf:
> @lulf great news! Insofar as the spec, I recall somewhere it being said "the device SHALL send a Join Request on DR0" so I defaulted to 0. I admit, I'm having trouble finding the reference right now.
> 
> In theory, I think low datarates are considered most conservative since spreading factor is highest. Of course, this causes issues when either lots of collisions are possible or when oscillators have trouble maintaining a precise timing over such a long transmit...
> 
> Anyway, I'm open to other approaches long-term. Insofar as this PR, I think leaving a setter accessible can allow for customization by the user.
> 

Great. Would it be better to expose an enum selecting the rate rather than the magic array index?

> Also, FYI, I'm working on a quick README update which, amongst other things, will reference your project if that's OK?

Thats fine with me, thanks!

lthiery:
> 
> Great. Would it be better to expose an enum selecting the rate rather than the magic array index?
> 

It absolutely would be but TBH I didn't know enough about datarates to type it properly. I think if we do:

```
enum DR {
_0
..15
}
```

that should suffice for now?

lulf:
@lthiery Ah, I was thinking about configuring bandwidth/spreadingfactor to select the datarate, but if that doesn't make sense (I don't know anything about data rates :)), just leave it as is.

lthiery:
Wow, check this beauty out:

image

Given this, I do believe I will set default data rate as another part of the RegionHander trait.

> @lthiery Ah, I was thinking about configuring bandwidth/spreadingfactor to select the datarate, but if that doesn't make sense (I don't know anything about data rates :)), just leave it as is.

Generally, I like to: (1) make an API safe, (2) reduce breaking API changes. So I think I'll just do it as I spec'ed previously as it will be a step in the right direction. There will perhaps be a smarter way to do it later when we or someone has a better idea.

lthiery:
Ok, I'm calling this done unless there are any concerns @lulf @ivajloip

lthiery:
I was trying to be more deliberate and make some of the region types public only to the crate and I realized I introduced some issues for clients which had previous demanded the RegionHandler be public. I did some work decoupling what needed to be private (ie: the region::Configuration struct) from what could be private (ie: Frame, Window, RegionHandler).

lthiery:
> I did a somewhat quick review of the code and generally it looks good to me. Thank you very much for helping in the development of a Rust LoRaWAN stack! It sounds like all issues have been resolved and now we can merge this?

Yes, I think we're good to go as @lulf initial trouble with Joining seemed not associated with a code regression but instead a different datarate used on Join.

PR #34, opened 2021-03-24
Update to newer versions of aes and cmac crates

lulf:
This reduces memory usage of generating keys when joining LoRa networks,
which is useful on embedded devices.

### DIFF
--- a/encoding/Cargo.toml
+++ b/encoding/Cargo.toml
@@ -10,8 +10,8 @@ keywords = ["lorawan", "iot", "lpwan", "parser", "lightweight"]
 readme = "README.md"
 
 [dependencies]
-aes = { version = "0.4.0", optional = true }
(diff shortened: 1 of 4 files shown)


ivajloip:
That seems quite nice, thanks! Running `cargo clippy --all-targets --all-features -- -D warnings ` I see that the benchmarks also need some updating, could you do that? If not I will do it this weekend and merge the change.

lulf:
@ivajloip Sorry about that, it should be fixed now. FWIW I got a lot of warnings from clippy that was seemingly unrelated, maybe I have a different version.

ivajloip:
Thanks! I will check clippy's output, but having new things to fix does happen sometimes with newer stable version of the toolchain, so I guess it's time to update mine :slightly_smiling_face:

PR #36, opened 2021-04-22
Make receive delay1 configurable

lulf:
Prototype on making receive_delay1 configurable. If you think this is an acceptable approach, I can adjust the settings for the other delays to make them configurable as well.

@ivajloip @lthiery

### DIFF
--- a/device/src/region/mod.rs
+++ b/device/src/region/mod.rs
@@ -132,7 +132,7 @@ impl Configuration {
         }
     }
 
-    // RECEIVE_DELAY2 is not configurable, LoRaWAN 1.0.3 Section 5.7: "The second reception slot opens one second 840after the first reception slot."
+    // RECEIVE_DELAY2 is not configurable, LoRaWAN 1.0.3 Section 5.7: "The second reception slot opens one second after the first reception slot."
     pub fn set_receive_delay1(&mut self, delay: u32) {
         self.receive_delay1 = delay;
         self.receive_delay2 = self.receive_delay1 + 1000;
--- a/device/src/region/mod.rs
+++ b/device/src/region/mod.rs
@@ -16,6 +16,10 @@ pub use us915::US915;
 
 pub struct Configuration {
     state: State,
+    join_accept_delay1: u32,
+    join_accept_delay2: u32,
+    receive_delay1: u32,
+    receive_delay2: u32,
 }
 
 // This datarate type is public to the device client
@@ -69,12 +73,12 @@ pub(crate) struct Datarate {
     spreading_factor: SpreadingFactor,
 }
 
-pub(crate)  enum Frame {
+pub(crate) enum Frame {
     Join,
     Data,
 }
 
-pub(crate)  enum Window {
+pub(crate) enum Window {
     _1,
     _2,
 }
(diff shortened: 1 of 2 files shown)


lthiery (changes requested):
This approach looks generally good to me. 

In particular, I realize now that putting `rx_delay` in the RegionHandler definition was wrong, as those are supposed to be "region specific", IMO. Meanwhile, the RECEIVE_DELAY1/2 are considered "out of band configuration variables" for an end-device, so they belong in that Configuration struct where you've put it.

That being said, rather than store "receive_delay1" in the struct, I'd encourage you to store all of the different delays, ie: JOIN_ACCEPT_DELAY1/2 and RECEIVE_DELAY1/2. In fact, without it, you'll have a problem with your TTN configuration: RECEIVE_DELAY2 (2) is prior to self.receive_delay1 (5).

device/src/region/mod.rs
```diff
@@ -178,7 +178,16 @@ impl Configuration {
...
     pub(crate) fn get_rx_delay(&self, frame: &Frame, window: &Window) -> u32 {
-        region_dispatch!(self, get_rx_delay, frame, window)
+        match frame {
+            Frame::Join => match window {
+                Window::_1 => JOIN_ACCEPT_DELAY1,
+                Window::_2 => JOIN_ACCEPT_DELAY2,
+            },
+            Frame::Data => match window {
+                Window::_1 => self.receive_delay1,
+                Window::_2 => RECEIVE_DELAY2,
```
ivajloip, line 188: It appears here that RECEIVE_DELAY2 could be changed separately from receive_delay1. My reading of the 1.0.3 specification, section 5.7 `The second reception slot opens one second after the first reception slot.` suggest otherwise. Did I miss something? :slightly_smiling_face:

ivajloip (commented):
That seems fine as an approach to me as well. Only one point to clarify/fix and I think we can merge this.

device/src/region/mod.rs
lulf: Good point, I think we should add the increment as a constant instead and always add that. (And not make RECEIVE_DELAY2 configurable)

lulf:
@ivajloip @lthiery Thank you for the review. I've updated the PR to allow changing all settings except receive_delay2, which is automatically set to receive_delay1 + 1 second.

device/src/region/mod.rs
```diff
@@ -113,19 +117,36 @@ macro_rules! region_dispatch {
...
         }
     }
-    pub(crate) fn create_tx_config(
-        &mut self,
-        random: u8,
-        datarate: DR,
-        frame: &Frame,
-    ) -> TxConfig {
+
+    // RECEIVE_DELAY2 is not configurable, LoRaWAN 1.0.3 Section 5.7: "The second reception slot opens one second 840after the first reception slot."
```
ivajloip, line 135: nit: `840after` should at least be separated to `840 after`. Probably 840 is not necessary at all.

ivajloip (approved):
Looks good to me! Given that Louis has more experience with that part of the code, I would prefer to know his view as well before merging.

ivajloip:
Thanks, Ulf!

PR #39, opened 2021-05-03
Remove heapless from radio api

lulf:
This goes further than #38 and attempts to make the radio buffer type configurable.

This makes it easier to depend on different versions of heapless in crates that depend on rust-lorawan. This also eases migration to creates that support const generics, as it would not break dependencies.


ivajloip:
That makes sense, thanks! :slightly_smiling_face:

lthiery (commented):
Great improvement!

PR #40, opened 2021-06-29
ABP support

lthiery:
Implements support for Activation by Personalization (ABP).

### DIFF
--- a/device/README.md
+++ b/device/README.md
@@ -22,8 +22,9 @@ A state machine diagram is provided in `src/state_machines/session`
 The following LoRaWAN features are implemented:
 
 - Class A device behavior
+- Over-the-Air Activation (OTAA) and Activation by Personalization (ABP)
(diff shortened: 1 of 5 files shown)


lulf (commented):
Another approach could be to introduce a JoinMode enum (ABP, OTAA), did you consider it? Changes looks good, nice work!

ivajloip:
This looks good indeed! I am also curious if there is a particular reason this approach was chosen over using a JoinMode enum as Ulf has suggested.

lthiery:
I agree with the suggestion - I was hoping to fix it last week but things have gotten in they. Hoping to get to it soon!

lthiery:
@ivajloip I implemented the enum and did some rebasing. Let me know if there's anything else!

ivajloip:
Thank you so much, @lthiery !

PR #42, opened 2021-09-01
Remove associated buffer type from radio and pass TX buffer as slice

lulf:
This decouples the TX and RX buffer more clearly in the radio API:

* The TX buffer is now passed as a [u8] slice rather than using the
  associated radio type. This makes sense because the radio is only
  given access to this buffer in TxRequest event, in which case it is
  now handed a &[u8] rather than a custom type. It also now only has
  read access to the TX buffer.
* The RX buffer is accessed by lorawan-device after an RxDone event, in
  which case it really only needs a &mut [u8] to parse the packet, and
  -device doesn't really change at all.
* Introduce a RadioBuffer type thats used internally to wrap the
  provided TX buffer slice, with a way to conveniently append to the
  buffer.

There are lots of lifetime arguments being added, in order to allow the
TX buffer to be provided externally. I believe this will be useful in order to fix get_random to use 
proper traits as well. There are some important performance benefits from these changes:

* TX buffer allocation can be done externally to lorawan-device,
  which is crucial for keeping async future stack usage down because the
  size of the buffer will not impact the stack anymore.
* Radio implementations are now free to decide how the RX buffer is
  managed as well, for the same reasons as above.

Although I would ideally like TX and RX to share buffer, I think that would better come in a later PR.

### DIFF
--- a/device/src/radio/mod.rs
+++ b/device/src/radio/mod.rs
@@ -59,22 +59,22 @@ pub trait PhyRxTx {
 
 pub struct RadioBuffer<'a> {
     packet: &'a mut [u8],
-    len: usize,
+    pos: usize,
 }
 
 impl<'a> RadioBuffer<'a> {
     pub fn new(packet: &'a mut [u8]) -> Self {
-        Self { packet, len: 0 }
+        Self { packet, pos: 0 }
     }
 
     pub fn clear(&mut self) {
-        self.len = 0;
+        self.pos = 0;
     }
 
     pub fn extend_from_slice(&mut self, buf: &[u8]) -> Result<(), ()> {
-        if self.len + buf.len() < self.packet.len() {
-            self.packet[self.len..self.len + buf.len()].copy_from_slice(buf);
-            self.len += buf.len();
+        if self.pos + buf.len() < self.packet.len() {
+            self.packet[self.pos..self.pos + buf.len()].copy_from_slice(buf);
+            self.pos += buf.len();
             Ok(())
         } else {
             Err(())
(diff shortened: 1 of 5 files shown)


lthiery (approved):
This looks like a great change :+1:

device/src/radio/mod.rs
```diff
@@ -44,35 +43,53 @@ where
...
 
     // we require mutability so we may decrypt in place
-    fn get_received_packet(&mut self) -> &mut Self::PhyBuf;
+    fn get_received_packet(&mut self) -> &mut [u8];
     fn handle_event(&mut self, event: Event<Self>) -> Result<Response<Self>, Error<Self>>
     where
         Self: core::marker::Sized;
 }
+
+pub struct RadioBuffer<'a> {
```
lthiery, line 60: A much friendlier name than PhyRxTxBuf :D

device/src/radio/mod.rs
```diff
@@ -44,35 +43,53 @@ where
...
+    packet: &'a mut [u8],
+    len: usize,
+}
+
+impl<'a> RadioBuffer<'a> {
+    pub fn new(packet: &'a mut [u8]) -> Self {
+        Self { packet, len: 0 }
```
lthiery, line 67: Isn't `len` actually `pos`?

lthiery (changes requested):

device/src/radio/mod.rs
lulf: Yeah, i squinted a bit when reading that just now as well. Fixed

lthiery (approved):

ivajloip:
Yes, that seems very nice! Thank you so much! :slightly_smiling_face:

PR #43, opened 2021-09-10
Bump `heapless` version

jorgeig-space:
Bump `heapless` to 0.7.5 and remove the `typenum` uses in internal Vectors.

No real benefit, but better to keep up with the latest data structure versions to avoid surprises with the following ones!


jorgeig-space:
I just realized I did this all in the wrong order 😅 , sorry! If you're OK with the other 2 PRs and want this one too, I'll update it then.

ivajloip:
Thank you, @jorgeig-space, for your contribution! The other PRs also look good, I will look into joining them a bit later today.

PR #45, opened 2021-09-10
Add defmt feature

jorgeig-space:
This will make it easier to debug applications using `defmt`. It's implemented only for the main types in `device` and for the keys in `encoding`, which are the parts users see.


lthiery:
I hadn't come across `defmt` yet! Very neat

jorgeig-space:
> I hadn't come across `defmt` yet! Very neat

It's the best thing since sliced bread :) The memory footprint is minimal, and the wire protocol is also minimal, so it's amazing not only for debugging but also to produce diagnostic messages in production that you can then dump OTA.

PR #46, opened 2021-09-12
Allow negative window offset in no_session sm

jorgeig-space:
If a negative value is specified in `get_rx_window_offset_ms`, the `no_session` state machine will panic because it is converted to u32 and causes the sum to calculate the window delay to overflow. This PR brings the sum in line with how it's done in the `session` state machines so it supports negative offsets.


lulf (approved):
LGTM!

Issue #41, opened 2021-08-30
Allow async radio implementations


lulf:
With async-await rust gaining more usage in embedded, and mechanisms for async traits being feature complete on nightly, it's not too long before stable rust will support it. There are now async SPI implementations, which in turn will allow async radio drivers. The advantages of that would be to use less power awaiting radio irqs, polling for busy signals (on some radios) without explicitly handling those peripheral-specific states at the "link layer". 

I'd like to sketch up a proposal for an AsyncPhyRxTx trait, supported in nightly rust configured using a Cargo feature. The code that calls into the radio would need to be async as well, so I was thinking to introduce a new API/types for that, but keep the internals mostly shared apart from the radio interaction.

Wdyt @ivajloip @lthiery ?

lthiery:
That sounds really great! A lot of the current device driver design was a result  of me trying [to wrap in Rust the vendor radio drivers](https://github.com/lthiery/rust-sx12xx) while leaving seams for pure Rust drivers. I think that, with async-await gaining traction in embedded, a new radio API focused on enabling async-await usage would be fantastic.

> I'd like to sketch up a proposal for an AsyncPhyRxTx trait, supported in nightly rust configured using a Cargo feature. The code that calls into the radio would need to be async as well, so I was thinking to introduce a new API/types for that, but keep the internals mostly shared apart from the radio interaction.

I have vague notions on how this could work but I definitely haven't had a hands-on work with the embedded sync stuff so I'm very excited to see what you come up with.

ivajloip:
That sounds indeed very interesting! I would love to see this and if you need any discussion before diving in a complete implementation, let me (and I believe Louis) know.

lulf:
Glad to hear the openness to the idea! I have experimented a bit, and got a working async radio. I've needed to make some changes:

* Reduce stack usage of owned data in Session. This is fixed in #48

Changes I plan to implement next:

* Create async version of handle_event from Device, through Session/NoSession and for all the relevant. This is necessary in order to allow async-await inside the states. The non-async logic that doesn't call the radio can be moved into helper methods that can be called by both implementations.

* Add an 'async' feature gate that will enable the async features at compile time. This will enable the async versions of the handle_event methods, as well as define a PhyRxTx trait that allows radio implementations to use async as well. 

[...]

PR #49, opened 2021-09-29
Add asynchronous implementation of a LoRaWAN device

lulf:
This adds an asynchronous implementation of a LoRaWAN device, where the
state machine is driven by async-await blocks, and allowing for an
asynchronous radio implementation.

The existing 'event driven' device remains untouched, but some types is
reused in the asynchronous implementation.

@ivajloip @lthiery I decided to try out creating a fully async version with async-friendly APIs instead of reusing the state machine. This also made the implementation code very simple to read also IMHO.

I've verified this works with several of the examples in drogue-device and will be moving them to use this


lulf:
@ivajloip Would you mind merging this as well if it looks ok, as discussed in #41 ?

lulf:
FWIW, I implemented @lthiery's suggestion of sharing create_join_request on the Credentials type as well.

PR #51, opened 2021-10-25
Add in github actions CI

lthiery:


ivajloip (approved):
I imagine we should provide the token somewhere in the settings of the project - I am checking that, but otherwise that looks good to me. Thanks! :)

ivajloip:
From what I saw, the current settings should be fine for the actions to start running. Probably this code needs to be committed before we see the first runs (which makes sense I think).

lulf (commented):
The job should probably build with the 'async' feature as well

ivajloip:
> The job should probably build with the 'async' feature as well

That sounds as a good idea! Both addressed in this PR or in a separate one is fine for me.

PR #54, opened 2021-11-22
Add checks permission on GITHUB_TOKEN

ivajloip:
This is expected to fix automated builds on master.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,6 +5,10 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+    
+permissions:
+  checks: write
+  contents: read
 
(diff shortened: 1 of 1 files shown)


lthiery:
image
something still ain't right

lthiery:
image
Is there a panel in settings to control this? @ivajloip

ivajloip:
I guess I finally figured the expected value out. Apologies for the noise! AFAIU the UI allows only to allow way too many things or only read permission for content (+metadata which apparently is there by default), both of which were not fine. I hope that's the only missing permission! :slightly_smiling_face:

lthiery:
Give it a whirl on master then I guess!

lthiery (approved):

PR #56, opened 2021-12-07
Improve async radio trait to allow non-static impls

lulf:

PR #57, opened 2021-12-20
Fixes for clippy

lthiery:


lulf:
LGTM, but I think these are already part of https://github.com/ivajloip/rust-lorawan/pull/56  as well, which needs @ivajloip's review before it can be merged.

lulf (approved):

lthiery:
Yeah I saw but I was hoping to break out the clippy fixes so we can land it ASAP. But if yours lands first, let's close this one.

ivajloip:
My understanding is that this is now merged as part of #56 , please reopen if that is not the case.

PR #58, opened 2021-12-20
CFList with types

lthiery:


lthiery:
Since the occurrence of 1.0.4, there are now two types of CfList: 
* Dynamic Frequency, which has always excited
* Fixed Frequency, which is new
The highest byte of CfList indicates the type.

ivajloip (approved):
Thanks for the contribution! Please be sure to squash the two commits

PR #59, opened 2021-12-22
Switch to Const Generic for Transmit Buffer

lthiery:


lulf (approved):
You're forced to stack allocate the buffer with this change, but it a not that big of a problem now that the async impl is separate, and it's what I care about personally :)

lthiery:
I personally don't see a major downside to that. Why would it be a problem in your view? Someone can always stuff it in an option if they want to put the whole Device struct in a static.

Maybe I'll work on another PR to make fn new a const.

lthiery:
> Maybe I'll work on another PR to make fn new a const.

Ah - `trait bounds other than `Sized` on const fn parameters are unstable`. So I guess that's not an option for now

lulf:
> I personally don't see a major downside to that. Why would it be a problem in your view? Someone can always stuff it in an option if they want to put the whole Device struct in a static.
> 

Yeah, that's certainly doable, so I wouldn't worry about it either.

> Maybe I'll work on another PR to make fn new a const.

One common approach is to use UnsafeCell or MaybeUninit for it, so I think it's fine without const new

PR #60, opened 2021-12-22
implement RxDelay handling from Join and MAC Commands

lthiery:
RxDelay may be configured in the JoinAccept or in a MAC Command attached to a downlink.

This PR parses RxDelay from either of these frame types and saves the delay timing. It enqueues the RXTimingSetupAns for the next uplink.

### DIFF
--- a/device/src/mac.rs
+++ b/device/src/mac.rs
@@ -6,9 +6,7 @@ During Uplink assembly, this struct will be inquired to drive construction
 use heapless::Vec;
 
 use super::region;
-use lorawan_encoding::maccommands::{
-    LinkADRAnsPayload, MacCommand, RXTimingSetupAnsPayload, RXTimingSetupReqPayload,
-};
+use lorawan_encoding::maccommands::{LinkADRAnsPayload, MacCommand, RXTimingSetupAnsPayload};
 
 #[derive(Default, Debug)]
 pub struct Mac {
--- a/device/src/mac.rs
+++ b/device/src/mac.rs
@@ -18,13 +18,19 @@ type AdrAns = u8;
 type RxDelayAns = bool;
 
 //work around for E0390
-trait AdrAnsTrait {
+trait MacAnsTrait {
     fn add(&mut self);
     fn clear(&mut self);
+    // we use a uint instead of bool because some ADR responses
+    // require a counter for state.
+    // eg: ADR Req may be batched in a single downlink and require
+    // multiple ADR Ans in the next uplink
     fn get(&self) -> u8;
 }
 
-impl AdrAnsTrait for AdrAns {
+// multiple AdrAns may happen per downlink
+// so we aggregate how many AdrReqs are reqired
+impl MacAnsTrait for AdrAns {
     fn add(&mut self) {
         *self += 1;
     }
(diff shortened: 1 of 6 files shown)


device/src/state_machines/session.rs
```diff
@@ -414,6 +414,7 @@ impl WaitingForRx {
                             if let Ok(PhyPayload::Data(DataPayload::Encrypted(encrypted_data))) =
                                 lorawan_parse(shared.radio.get_received_packet(), C::default())
                             {
+                                let confirmed = encrypted_data.is_confirmed();
```
ivajloip, line 417: I failed to notice how this is going to be used. Did you mean `self.confirmed`?

device/src/region/mod.rs
```diff
@@ -188,8 +189,11 @@ impl Configuration {
 
     pub(crate) fn process_join_accept<T: core::convert::AsRef<[u8]>, C>(
         &mut self,
+        mac: &mut crate::Mac,
```
ivajloip, line 192: nit: I have a slight preference of having `mac` as the second parameter, as my perception is that `join_accept` is usually the first parameter when provided and if that's the case, it's worth keeping the consistency.

device/src/mac.rs
```diff
@@ -6,20 +6,24 @@ During Uplink assembly, this struct will be inquired to drive construction
...
 pub struct Mac {
     adr_ans: AdrAns,
+    rx_delay_ans: RxDelayAns,
 }
 
 type AdrAns = u8;
+type RxDelayAns = bool;
 
 //work around for E0390
 trait AdrAnsTrait {
```
ivajloip, line 23: does that name still reflect well the way it's used?

device/src/mac.rs
```diff
@@ -6,20 +6,24 @@ During Uplink assembly, this struct will be inquired to drive construction
...
     fn add(&mut self);
     fn clear(&mut self);
-    fn get(&mut self) -> u8;
+    fn get(&self) -> u8;
```
ivajloip, line 26: nit: could you add a short comment maybe on the type level what it's used for? It might make reading the code slightly easier next time.

ivajloip (dismissed):

device/src/state_machines/session.rs
lthiery: This was a mistake. Carry over from the confirmed uplink PR

device/src/mac.rs
lthiery: I added some comments on the Trait impl which hopefully clarifies

device/src/mac.rs
```diff
@@ -6,49 +6,93 @@ During Uplink assembly, this struct will be inquired to drive construction
...
+use lorawan_encoding::maccommands::{LinkADRAnsPayload, MacCommand, RXTimingSetupAnsPayload};
 
 #[derive(Default, Debug)]
 pub struct Mac {
     adr_ans: AdrAns,
+    rx_delay_ans: RxDelayAns,
 }
 
+// multiple AdrAns may happen per downlink
+// so we aggregate how many AdrReqs are required
```
lthiery, line 18:
```suggestion
// so we aggregate how many AdrAns are required
```

ivajloip (approved):

PR #61, opened 2021-12-22
ACK Confirmed Downlinks

lthiery:


encoding/src/parser.rs
```diff
@@ -970,6 +975,14 @@ impl<'a> FHDR<'a> {
 pub struct FCtrl(u8, bool);
 
 impl FCtrl {
+    pub fn new_uplink() -> FCtrl {
+        FCtrl(0x0, true)
+    }
+
+    pub fn set_ack(&mut self) {
```
ivajloip, line 982: While this is a tiny type, so I can see the logic in suggesting to have this method here, but still I would rather have a separate builder for FCtrl and have that one used in `device/src/state_machines/session.rs` as it will be more consistent and easier to use when we need to set multiple properties of the type.

device/src/mac.rs
```diff
@@ -35,6 +36,16 @@ impl AdrAnsTrait for AdrAns {
 }
 
 impl Mac {
+    pub fn set_confirmed(&mut self) {
+        self.confirmed = true;
+    }
+
+    pub fn is_confirmed(&mut self) -> bool {
```
ivajloip, line 43: it's unexpected for me to see changes to state in an `is_` method. In order to reduce risk of future confusion, I think it should be called something more complicated or split into two methods.

ivajloip (changes requested):

lulf (approved):
LGTM!

ivajloip (approved):

lthiery:
I had to do some rebasing due to other branches that landed so I required a new +1

lulf (approved):

ivajloip (dismissed):

lthiery:
Rebased again! But i think we're good now. Almost all PR's are through

ivajloip (approved):

PR #72, opened 2022-05-29
Adding basic support for AU915

bldh:


lthiery (changes requested):
@bldh Thank you for adding AU915 support!

Please excuse a few nits, as I may be pointing out things that are wrong in AU915 that _I_ probably also did wrong in US915. But I know the regional specification a lot better now then I did back then and will be hopefully revisiting this area of the code soon.

lthiery:
@bldh any interest in completing this? i can pick it up from here, if not.

lthiery:
replaced with #81

PR #78, opened 2022-07-18
Fix RX2 window delay

timokroeger:
intermediate fix for #77

PR #79, opened 2022-08-03
Async rx offset based

lulf:
Based upon #78 - but changing to use an offset based calculation. Also use correct frame parameter for calculating delays.

@timokroeger Its still not working, but I think this change might be worthwhile anyway. I'll try it with a different radio than thw wl5 to check if it works.

### DIFF
--- a/device/src/async_device/mod.rs
+++ b/device/src/async_device/mod.rs
@@ -137,7 +137,8 @@ where
                             Err(Error::InvalidMic)
                         }
                     }
-                    _ => Err(Error::UnableToDecodePayload("")),
+                    Err(err) => Err(Error::UnableToDecodePayload(err)),
+                    _ =>  Err(Error::UnableToDecodePayload("")),
                 }
             }
             JoinMode::ABP {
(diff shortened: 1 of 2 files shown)


timokroeger:
I tested your `Fixes` commit on a LoRA-E5 Dev Board and I was able to join a LoRaWAN network.
Did you implement the offset based traits for `embassy-lora` already?

lulf:
@timokroeger I have them implemented on this branch https://github.com/lulf/embassy/tree/debug

lthiery:
@lulf any idea what this nightly fail is? why can't it find defmt?

lulf:
@timokroeger I'll try to adopt the radio changes to your branch and test again. Also I have a lora-e5 to test with, as well as a sx127x radio which could indicate where the issue is.

lulf:
@lthiery It is just some debug print statements I've put into the code, and iirc nightly compiles without defmt logging enabled (as it's an optional dependency), which is why it fails. I'll fix the statements before making it ready for review.

lulf:
I found the reason for the bad packet decode error that I got. It was due to a missing self.radio_buffer.clear() in these PRs. I also had to tune the sx127x radio rx_window_duration to a larger value in order to make RX successful, but iirc the values I've used might have been influenced by the bug.

Unrelated to this PR, but I'm still not able to join using your lora-e5 PR with embassy @timokroeger (only with master + timings adjustments) so I will need to look into that.

CBJamo:
I can confirm that this code works with my hardware (sx1262 connected directly to a RPI). I had to make some small tweaks related to the output type of the timer I'm using, but that shouldn't do anything to the timing logic. 

The timing logic is much cleaner as well, this looks good to me.

lulf:
@ivajloip or @lthiery could one of you have a look?

lthiery (approved):

PR #80, opened 2022-08-18
Clippy fixes and version bump for device

lulf:


lulf:
I'd like to release a new version with lorawan-device after the API changes for the async radio.

ivajloip (approved):

PR #83, opened 2022-09-12
Cap RX1 window duration so that RX2 can start on time

lucasgranberg:
If the timings are set to 3000ms window open time the RX2 window can't start at 2000ms.
Suggest adding cap to the RX1 window to account for this.


lulf (approved):
Looks good to me, thanks!

PR #85, opened 2022-09-16
Add AS923 and IN865 regions

CBJamo:
This is an implementation of AS923 (including subbands) and IN865 regions. 

For AS923_1, there is also AS923_1B. As I understand it, there is no difference on a device, only on gateways, but I'd appreciate a double check on that.


lthiery (commented):
This looks right to me. I think what you're starting to do with AS923 shows how the regions could eventually be handled as "Dynamic" or "Fixed" frequency plans which would help make this code less repetitive.

CBJamo:
Closing this in favor of #96

PR #86, opened 2022-09-29
Allow SessionData to be extracted from Device

lucasgranberg:
Right now there is no way to persist the SessionData to flash. This is a requirement for LoraWan 1.0.4 but is also nice to have just for being able to delete the Device for saving power in sleep.

### DIFF
--- a/encoding/Cargo.toml
+++ b/encoding/Cargo.toml
@@ -14,7 +14,7 @@ aes = { version = "0.6.0", optional = true }
 cmac = { version = "0.5.1", optional = true }
 generic-array = "0.14.4"
 defmt = { version = "0.3.0", optional = true }
-serde = { version = "1.0.145", default-features = false, features = ["derive"]}
+serde = { version = "1.0.145", default-features = false, features = ["derive"], optional = true}
 
 [dev-dependencies]
 criterion = "0.3"
@@ -26,11 +26,8 @@ harness = false
 
 [features]
 default = ["full"]
-
 full = ["with-to-string", "with-downlink", "default-crypto"]
-
 default-crypto = ["aes", "cmac"]
-
 with-to-string = []
-
 with-downlink = []
+serde = ["dep:serde"]
--- a/encoding/Cargo.toml
+++ b/encoding/Cargo.toml
@@ -14,6 +14,7 @@ aes = { version = "0.6.0", optional = true }
 cmac = { version = "0.5.1", optional = true }
 generic-array = "0.14.4"
 defmt = { version = "0.3.0", optional = true }
+serde = { version = "1.0.145", default-features = false, features = ["derive"]}
 
 [dev-dependencies]
 criterion = "0.3"
--- a/encoding/src/keys.rs
+++ b/encoding/src/keys.rs
@@ -6,11 +6,11 @@
 //
 // author: Ivaylo Petrov <ivajloip@gmail.com>
 use super::securityhelpers::generic_array::{typenum::U16, GenericArray};
-use serde::{Deserialize, Serialize};
 
 /// AES128 represents 128 bit AES key.
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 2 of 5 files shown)


lulf (commented):
Great work, I think this is a good start to re-use session data. As discussed in chat:

* Change fn get_session() to return a ref
* Remove derive Copy
* Probably want to test ser/deser of the SessionData in a unit test

lulf (dismissed):
Looks good to me!

lthiery (approved):

lthiery:
Excellent! Been wanting this feature.

encoding/src/parser.rs
```diff
@@ -44,7 +45,7 @@ macro_rules! fixed_len_struct {
         struct $type:ident[$size:expr];
     ) => {
         $(#[$outer])*
-        #[derive(Debug, Eq)]
+        #[derive(Debug, Eq, Serialize, Deserialize)]
```
ivajloip, line 48: the same as above.

encoding/src/keys.rs
```diff
@@ -6,10 +6,11 @@
 //
 // author: Ivaylo Petrov <ivajloip@gmail.com>
 use super::securityhelpers::generic_array::{typenum::U16, GenericArray};
+use serde::{Deserialize, Serialize};
 
 /// AES128 represents 128 bit AES key.
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
```
ivajloip, line 13: I might be misunderstanding the intend here, but wouldn't that require serde to build? I think this is a great default, but I am curious how confident can we be that this is always what people will want? Maybe we can provide a new feature and two variants - if the feature is enabled or if it's not? That way users can disable it if they so desire (I can imagine for example gateway that partially parses data and logs something or device that uses an unusual way to store the session might not want this).

encoding/Cargo.toml
```diff
@@ -14,6 +14,7 @@ aes = { version = "0.6.0", optional = true }
 cmac = { version = "0.5.1", optional = true }
 generic-array = "0.14.4"
 defmt = { version = "0.3.0", optional = true }
+serde = { version = "1.0.145", default-features = false, features = ["derive"]}
```
ivajloip, line 17: Maybe that should be optional (see below),

ivajloip (changes requested):
Hi,

That seems really useful! Thank you so much for the pull request! Could you confirm if my understanding is correct (see below) and possibly explain me what I missed? :slightly_smiling_face:

encoding/src/keys.rs
lucasgranberg: It could put it behind a feature. Alternatively we could let the user impl it themselves but that is also cumbersome.

I was planning on using Postcard for Serde to get the SessionData into bytes that can be saved to flash but it uses varint so it is variable length. Maybe we should skip serde alltogether and just use `core::slice::raw_from_parts`. Usually the user doesn't really need the SessionData into anything else than bytes. They could impl serde themselves if they need to.

lucasgranberg:
I made serde optional now. The reason behind even including it is to make it easier to save the session to flash. ie. we need a way to convert the SessionData struct to bytes.

In the future we might even include code for that into the repo as it is an requirement for LoraWan 1.0.4 to save the session even during power off and reboot.

encoding/src/keys.rs
lulf: I think making serde an optional feature is fine. You can still use fixed length records with postcard, as long as you take into account the max size and fill the rest with zeros. Or, you can use a different encoding that doesn't use varint.

core::slice::raw_from_parts will be compiler and target dependent, so I don't think that's something users should rely on.

lulf (approved):

ivajloip:
Thank you!

PR #92, opened 2022-10-14
add get_region to Device and make the region cloneable

lucasgranberg:
This would enable the user to drop the device and re-create it later. This works the same way as get_session but the session data doesn't include all fields that should be persisted.

https://github.com/ivajloip/rust-lorawan/issues/90


lulf:
Looks good to me. Any other changes you'd like to do since it's still a draft?

lucasgranberg:
Ill add serde stuff to region so that it can be serialized to flash if needed.

lulf (approved):

Issue #91, opened 2022-10-13
channel_enabled appears to index ChannelMask incorrectly


bldh:
Hi,

I have been having a look at implementing LinkADRReq for AU915 and have noticed that the channel_enabled appears to index the ChannelMask incorrectly. 

LoRaMAC-node stores channel masks directly from the LinkADRReq MAC command and when setting the 1st sub band using a ChMaskCtrl of 5 they write to the right most bits of the ChannelMask. [Link to example.](https://github.com/Lora-net/LoRaMac-node/blob/e851b079c82ba1bcf3f4d291ab69a571b0bf458a/src/mac/region/RegionAU915.c#L618)

Currently channel_enabled indexes as though index 0 stores the lowest channels of the mask.

I have implemented a potential fix for this [here](https://github.com/bldh/rust-lorawan/commit/e93296807cd81703e0ae0cbcdb09d97531f426bf)

lthiery:
@bldh thanks for bringing up the topic.

However, I think that your fix is not quite right. From the 1.0.4 specification,
```
The channel mask control ( ChMaskCntl ) field controls the interpretation of the previously defined ChMask bit mask. It controls the block of 16 channels to which the ChMask applies. It can also be used to turn all channels on or off globally using a specific modulation. The meaning of ChMaskCntl is region-specific and defined in the “LoRaWAN Regional Parameters” [RP002] document.
```

This tells me that we need to hook into region handlers with the ChMaskCntl field (contained in the Redundancy byte) rather than makes changes to the global ChannelMask encoding functions. Glancing at the regional specification, it seems the current code is correct for "Dynamic Channel Plans" (eg: EU868 or AS923) but "Dynamic Channel Plans" (eg: US915 and AU915) need special handling of the ChMaskCntl field when parsing the ChannelMask in LinkADR.

PR #95, opened 2022-12-27
make ChannelMask type with a generic const parameter

lthiery:
Using the recent const generic feature, we can make the ChannelMask type more generally useful. This makes the type work well when receiving ChannelMask in CFList from the JoinAccept in US915 and AU915; that is to say, asking the type if a specific channel is enabled works (but no effort in this PR is made for storing it and using it in the region handlers).

Note that #91 raises the issue of ChMaskCntl which this does not address. More work is needed for Fixed Channel Plans (US915 and AU915) should they leverage the ChannelMask in LinkADRReq; they currently do not.

### DIFF
--- a/encoding/src/maccommands.rs
+++ b/encoding/src/maccommands.rs
@@ -382,8 +382,8 @@ impl<'a> LinkADRReqPayload<'a> {
     }
 
     /// Usable channels for next transmissions.
-    pub fn channel_mask(&self) -> ChannelMask {
-        ChannelMask::new_from_raw(&self.0[1..3])
+    pub fn channel_mask(&self) -> ChannelMask<2> {
+        ChannelMask::<2>::new_from_raw(&self.0[1..3])
     }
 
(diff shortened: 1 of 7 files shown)


lulf (dismissed):
LGTM, nice improvement!

ivajloip (dismissed):
Thanks!

lulf (dismissed):

encoding/src/maccommands.rs
```diff
@@ -421,29 +422,29 @@ impl ChannelMask {
 
     /// Verifies if a given channel is enabled.
     pub fn is_enabled(&self, index: usize) -> Result<bool, &str> {
-        if index > 15 {
+        if index > N * 8 - 1 {
             return Err("index should be between 0 and 15");
```
lthiery, line 426:
```suggestion
            return Err("index should be between 0 and {}", N*8-1);
```

lulf (approved):

PR #96, opened 2022-12-29
Channel Plan Refactor

lthiery:
I've wanted to reorganize how the region code works to leverage the fact that there are two groups of channel plans that have a lot in common: Fixed Channel Plans (US915 & AU915) and Dynamic Channel Plans (the rest, such as EU868 and AS923).

This initial draft PR shows a proof of concept of how it would work for Fixed Channel Plans. With this refactor common logic between the two plans is shared in the `impl RegionHandler for FixedChannelPlan<F: FixedChannelRegion>`, while a new trait `FixedChannelRegion` is implemented by the US915 and AU915 modules. 

A few improvements are also made:
- CfList in JoinAccept from the 1.0.4 spec is leveraged to set channel mask
- The out of spec "set_subband" function is removed
- the concept of "subbands" is removed from the channel map (ie: array of 9 channel arrays if flattened to 72 channels); special care still needs to be taken between 500 kHz channels (64-71) and the rest, but that is managed elsewhere

I'd still like to develop something similar for the Dynamic Channel Plans, but wanted to share the idea for initial review.


lthiery:
Dynamic plans are refactored in the same fashion now and I added support for AS923-1, AS923-2, AS923-3, AS923-4. 

CN470 is a bit of a mess... I think when I implemented it I accidentally merged the old fixed plan and newer dynamic plan. With this PR, I will try to re-implement both styles appropriately.

lthiery:
I removed support for CN470 until I can figure out what to do with it. I've tested this PR on US915 and EU868 and it appears to work well! Ready for review.

lthiery:
@lulf I'm not quite sure what to do about the randomness API. I've changed things slightly on my end so that the region handlers can actually get as much entropy as needed (they get the function pointer instead of just 32-bits of entropy). I see you require RngCore to be implemented which is a much broader trait than necessary.

What're your thoughts about standardizing on just `fn get_random() -> u32`? That is to say, do you feel strongly about the trait requirement?

lulf:
Using `fn get_random() -> u32` is not a good approach for embedded rust, because it requires to define a static global with mutual exclusion in order to create safe rust code when using peripheral RNGs. Using a closure parameterized type `(F: Fn() -> u32)` is an alternative if you want the 'minimal trait', but once you need to pass a generic, it might as well be RngCore I think. The advantage is that all embedded random number generators implement RngCore as well as a lot of STD random number generators.

In addition, if you need cryptographically secure RNG, the API should actually require RngCore + CryptoRng, because that adds another assurance that the user is passing a proper secure RNG. 

That being said, maybe the API can be structured a bit differently so that you don't need to pass the rng all the way into the region handler, but just a u32 that can be 'u32 % range' inside the handler?

lthiery:
@lulf Thanks for the response. I see what you mean about it requiring a static global now, yeah. I'll copy your RngCore approach in the sync stack and integrate into this PR.

Insofar as passing the function all the way down, it's because getting a random number in a range that's actually random is kind of difficult due to [modulo bias](https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/).

So there is some amount of "re-sampling" of random numbers that I need to do to get a reasonably random number in range.

That being said, I should be able to pass the RngCore implementor by reference for the regional plans to use.

lulf (approved):
Quite a big chunk of work, thanks for doing it! 🚀

lthiery:
> Quite a big chunk of work, thanks for doing it! rocket

Np! It was festering in the back of my mind lately as implementing more of the spec really seemed tedious with this kind of refactor.

lulf (dismissed):

jbeaurivage:
@lthiery I've just managed to find the time to look at this. Thanks for putting all the work! Seems to work quite well on my end.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -143,7 +143,7 @@ impl<const D: usize, F: FixedChannelRegion<D>> RegionHandler for FixedChannelPla
     }
 
     fn get_rx_frequency(&self, _frame: &Frame, window: &Window) -> u32 {
-        let channel = self.last_tx_channel % 7;
+        let channel = self.last_tx_channel % 8;
```
lthiery, line 146: woops! nice catch

device/src/async_device/mod.rs
```diff
@@ -221,6 +221,13 @@ where
             .region
             .create_tx_config(&mut self.rng, self.datarate, &Frame::Data);
 
+        // Unless the same frame is to be retransmitted (see NbTrans parameter of LinkADRReq command, LoRaWAN spec
+        // 1.0.2 section 5.2 for retransmissions), FCnt must be incremented on each transmission.
+        self.session
```
lthiery, line 226: FWIW, in the sync stack, we wait increment FCnt in the "WaitingForRx" state:

- If a packet is received, we always increment the FCntUp; if this was a confirmed or unconfirmed frame, no matter.
- If the Rx2 window expires and it was an unconfirmed uplink, then we also increment.

So the issue with this implementation is that if it's a confirmed uplink with no confirmation, you've still incremented the FCntUp. This might make it difficult to implement retry on lack of confirmation, for example.

ivajloip (dismissed):
I finally managed to find the time to finish reviewing this. Thanks a lot for all the work you have put into this! And apologies for the delay! Also feel free to ping me on email in the future. :slightly_smiling_face:

encoding/src/maccommands.rs
```diff
@@ -421,29 +492,30 @@ impl ChannelMask {
...
+            return Err("index should be between 0 and {index_limit}");
         }
         Ok(self.channel_enabled(index))
     }
 
     /// Provides information for each of the 16 channels if they are enabled.
-    pub fn statuses(&self) -> [bool; 16] {
-        let mut res = [false; 16];
+    pub fn statuses<const M: usize>(&self) -> [bool; M] {
+        let mut res = [false; M];
```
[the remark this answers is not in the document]
lthiery: I can't think of any, but unfortunately:
```
error: generic parameters may not be used in const operations
   --> encoding/src/maccommands.rs:504:31
    |
504 |         let mut res = [false; N*2];
    |                               ^ cannot perform const operation using `N`
    |
    = help: const parameters may only be used as standalone arguments, i.e. `N`
```

PR #97, opened 2023-01-09
Clippy Clean up

lthiery:
Clippy gets smarter! Also, nightly fixed

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -54,7 +54,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-04-24
+          toolchain: nightly
           default: true
           override: true
           components: clippy, rustfmt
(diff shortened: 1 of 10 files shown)


.github/workflows/rust.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-04-24
+          toolchain: nightly
```
lulf, line 57: Using the latest nightly has a tendency to break unexpectedly, so I would recommend pinning just to a newer nightly version.

lthiery: It actually looks like the async stack no longer requires nightly. Maybe we should just do an all-features build from stable and call it a day?

lulf (approved):

ivajloip (approved):

.github/workflows/rust.yml
ivajloip: If we can switch to stable, that would be great I think. :)

ivajloip: If we can switch to stable, that would be great I think. :)

lulf: Good point, if it works with stable that is better.

PR #99, opened 2023-03-17
Fix RX1 channel calculation for fixed frequency plans

jbeaurivage:
There seems to be a bug in the RX1 channel calculation. According to the regional parameters document, the downlink channel for the RX1 window should be `(uplink channel % 8)`.

Screenshot_20230317_173651
Screenshot_20230317_173633


lulf (approved):
Looks good, thanks!

PR #102, opened 2023-03-30
Allow the LoRa radio to implement `RngCore`

### LINKED ISSUE #101: Support chips with integrated RNG
There are many microcontrollers which don't have an onboard TRNG. On the other hand, some LoRa chips (eg, SX1262) can be used to generate random numbers. It would be nice to leverage that functionality and give the user the option of using either the microcrontroller (if it exists) or the LoRa chip's.

jbeaurivage:
Closes #101.

This PR allows the user to choose whether the LoRa chip implements `RngCore`, or provide an external RNG. This is useful for setups where the main uC doesn't have access to an RNG, but the LoRa chip provides one. One drawback is that the `RngCore` implementation is blocking, and therefore any async transactions (SPI, I2C, etc) will have to be done in a blocking manner to get random numbers, perhaps using `futures::executor::block_on`.


device/src/async_device/mod.rs
```diff
@@ -22,28 +22,74 @@ use lorawan::{
...
 
 type DevNonce = lorawan::parser::DevNonce<[u8; 2]>;
-use crate::radio::types::RadioBuffer;
 pub use crate::region::DR;
+use crate::{private::Sealed, radio::types::RadioBuffer, GetRng};
 pub mod radio;
 use core::cmp::min;
 
+/// Type-level version of the [`None`] variant
+pub struct NoneT;
```
lulf, line 31: What about just using the unit type '()'  for this?

lulf (commented):
Thanks, I like the approach!

lulf:
Regarding the blocking, you can perform blocking in async code, and I don't think it's a problem, and there is no async rng traits yet.

device/src/async_device/mod.rs
jbeaurivage: Sure, I can change it back to`()` if you prefer, obviously functionally it makes no difference. We use `NoneT` over at atsamd-hal, admittedly because the code is very heavy on type-level stuff. However, to me, looking at something like `Device<C, R, T, NoneT, N>` makes it instantly clear that the empty type is deliberate. Maybe it's just me, but `Device<C, R, T, (), N>` feels a little more ambiguous.

lulf: That's fair enough, I didn't think too much about that.

lulf (approved):

device/src/lib.rs
```diff
@@ -254,3 +254,17 @@ where
         result
     }
 }
+
+/// Trait used to mark types which can give out an exclusice reference to [`RngCore`].
+/// This trait is an implementation detail and should not be implemented outside this crate.
+#[doc(hidden)]
+pub trait GetRng: private::Sealed {
```
lthiery, line 261: pub(crate)?

lthiery (changes requested):
This looks great! A few thoughts.

1, It seems like the worst-case we want to avoid is where neither the regular Rng nor radio Rng is implemented. It seems like the constructor is the only way to do that. That's probably acceptable but it would be great to figure out how to enforce it at compile-time with types.

2. This PR doesn't seem to exercise the radio RNG as far as I can tell. Am I misreading that?

3. One suggestion for the Impls of the optional RNG trait on radios is to collect a bunch of random numbers at once and fill a buffer. Initiating a SPI txfer every time a random number is requested sounds expensive. You'll see in the code where random is used, sometimes we need a lot of random numbers at once (but we don't know ahead of time how many).

device/src/lib.rs
jbeaurivage: It's unfortunate, but `GetRng` has to be public, otherwise it's a compile error (can't leak private trait). That's why there's the `Sealed` trait, as a workaround to prevent other crates from providing their own implementations.

jbeaurivage:
> This looks great! A few thoughts.
> 
> 1. It seems like the worst-case we want to avoid is where neither the regular Rng nor radio Rng is implemented. It seems like the constructor is the only way to do that. That's probably acceptable but it would be great to figure out how to enforce it at compile-time with types.
> 
> 2. This PR doesn't seem to exercise the radio RNG as far as I can tell. Am I misreading that?
> 
> 3. One suggestion for the Impls of the optional RNG trait on radios is to collect a bunch of random numbers at once and fill a buffer. Initiating a SPI txfer every time a random number is requested sounds expensive. You'll see in the code where random is used, sometimes we need a lot of random numbers at once (but we don't know ahead of time how many).

Thanks for the review @lthiery. In response to your comments:

1. I currently don't know of a way to make the constraint of having at least one `RNG` provider enforceable at compile time *without* having two different constructors. Doing something like this
```rust
impl<R, C, T, const N: usize> Device<R, C, T, NoneT, N>
where
    R: radio::PhyRxTx + Timings + RngCore,
    C: CryptoFactory + Default,
    T: radio::Timer,
{
    /// Create a new instance of [`Device`] with a LoRa chip with a builtin RNG.
    /// This means that `radio` should implement [`rand_core::RngCore`].
    pub fn new(
        region: region::Configuration,
        radio: R,
        timer: T,
    ) -> Device<R, C, T, NoneT, N> {
        Device::new_with_session(region, radio, timer, NoneT, None)
    }
}

impl<R, C, T, G, const N: usize> Device<R, C, T, G, N>
where
    R: radio::PhyRxTx + Timings,
    C: CryptoFactory + Default,
    T: radio::Timer,
    G: RngCore,
{
    /// Create a new instance of [`Device`] with a RNG external to the LoRa chip.
    /// See also [`new_with_builtin_rng`](Self::new_with_builtin_rng)
    pub fn new(region: region::Configuration, radio: R, timer: T, rng: G) -> Device<R, C, T, G, N> {
        Device::new_with_session(region, radio, timer, rng, None)
    }
}
``` 
is a compile error, because as far as the compiler is concerned, these are two conflicting implementations of the same function. And it's kind of true: if you provided an external RNG *and* your radio implements `RngCore`, which implmentation should you choose? Having two constructors leaves the choice to the user.

2. The radio RNG is exercised when using `new_with_rng` because in that case, `Phy<R, NoneT>` implements `GetRng<RNG= R>`, therefore the `get_rng` method returns `&mut self.radio`. I can confirm that it works as I've tested the PR with a real device with a radio RNG and no TRNG onboard the MCU.

3. Having a buffer of random bytes pre-ready would probably work. I agree that a full SPI transaction (actually, it's multiple, at least in the case of SX1262). However a pre-filled buffer to me kind of sounds like kicking the can down the road, because if the device is left running long enough, we're bound to run out of random bytes eventually, unless the user can manually call a method that refreshes the randomness buffer at their discretion. Although tbh I'd rather not expose that implementation detail to the user... thoughts?

lthiery (approved):
Thanks for the response. Makes sense to me!

ivajloip (approved):

PR #103, opened 2023-04-08
Fix possible panic in parsing invalid packets

### LINKED ISSUE #94: Out of range panic while decoding packet
```
thread 'main' panicked at 'range end index 13 out of range for slice of length 12', /home/cbjamo/.cargo/git/checkouts/rust-lorawan-d3dc04aa7aafaad5/956bde4/encoding/src/parser.rs:565:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
 I had some debug logging running at the time, but unfortunately not RUST_BACKTRACE. Note that I'm a little behind the latest git, so the error is on line [parser.rs#L566](https://github.com/ivajloip/rust-lorawan/blob/master/encoding/src/parser.rs#L566) in the current source.

I believe the core issue is that when the device is parsing a packet (See [async](https://github.com/ivajloip/rust-lorawan/blob/master/device/src/async_device/mod.rs#L239) and [state machine](https://github.com/ivajloip/rust-lorawan/blob/master/device/src/state_machines/session.rs#L438)) to get the frame count to pass to the validate_mic function. We pass through the DataHeader trait's .fhdr() function, which, via the fhdr_length function, assumes the length field in the header is good.

So I got a bad downlink here, pulled some garbage out of the packet, and indexed out of bounds. I'm not sure what the cleanest way to fix this is. My gut says make the validate_mic function find the frame count on it's own, rather than passing through the trait's functions.

ivajloip:
Improve the validation of data packets length as incorrect validation can lead to panic.

Fixes #94.


encoding/src/parser.rs
```diff
@@ -627,9 +627,8 @@ impl<T: AsRef<[u8]>, F: CryptoFactory> EncryptedDataPayload<T, F> {
     }
 
     fn can_build_from(bytes: &[u8]) -> bool {
-        let has_acceptable_len = bytes.len() >= 12 &&
-            // TODO: Bug related to possibly insufficient number of bytes
-            fhdr_length(bytes[5]) <= bytes.len();
+        // The smallest packets contain MHDR (1 bytes) + FHDR + MIC (4 bytes).
+        let has_acceptable_len = 5 + fhdr_length(bytes[5]) <= bytes.len();
```
lthiery, line 631: Shouldn't we protect against bytes being less than 6 bytes before indexing into the 6th byte?

ivajloip: Initially I thought that as there is a protection already in the phy packet parsing logic, that would be enough, but given that people might want to use this code in other context, I agree that it makes more sense to have that protection here, so I added it + a test :)

lulf (approved):
LGTM!

PR #104, opened 2023-04-14
Add support for external LoRa physical layer functionality through the lora-phy crate

ceekdee:
Add another (optional) pathway to PhyRxTx/Timings support.

### DIFF
--- a/device/Cargo.toml
+++ b/device/Cargo.toml
@@ -21,11 +21,11 @@ defmt = { version = "0.3.0", optional = true }
 futures = { version = "0.3.17", default-features = false, optional = true }
 rand_core = { version = "0.6.2", default-features = false }
 serde = { version = "1.0.145", default-features = false, features = ["derive"], optional = true}
-lora = { version = "0.1.0", path = "../../lora", optional = true }
+lora-phy = { version = "0.1.0", path = "../../lora-phy", optional = true }
 
 [features]
 defmt = ["dep:defmt", "lorawan/defmt"]
 default = []
 async = ["futures"]
 serde = ["dep:serde"]
-external-lora = ["dep:lora"]
+external-lora-phy = ["dep:lora-phy"]
--- a/device/Cargo.toml
+++ b/device/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "lorawan-device"
-version = "0.9.0"
+version = "0.10.0"
 authors = ["Louis Thiery <louis@helium.com>", "Ulf Lilleengen <lulf@redhat.com>"]
 edition = "2018"
 categories = [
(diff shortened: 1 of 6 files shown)


ceekdee:
Updates need to be made when https://github.com/ivajloip/rust-lorawan/pull/105 is applied.

lulf:
I've merged the other PRs @ceekdee. What do you think about moving PhyRxTx to the lora-phy crate as suggested by @lthiery . With that change, this PR could implement only the Timings trait for the lora-phy types (maybe no need to introduce another type since it can implement the Timings trait directly on the lora-phy types?)

ceekdee:
Shouldn't rust-lorawan maintain control of how it uses the lora-phy API?  If PhyRxTx processing is placed in lora-phy, rust-lorawan loses some of that control.

Should lora-phy have a dependence on rust-lorawan?  That is, where is PhyRxTx declared?

lulf (dismissed):
After thinking on this a bit I agree with you @ceekdee, it does make more sense to keep the trait in rust-lorawan. Could you rebase the branch?

ceekdee:
Ready for review with updates for PhyRxTx.  It would be good if a new release can be done so I can use it when updating Embassy to support lora-phy.

ceekdee:
Just noticed that the Cargo.toml file for lorawan-device has:

[dependencies]
lorawan = { version = "0.7.1", path = "../encoding", default-features = false }  instead of 0.7.2

lulf (approved):
LGTM

ceekdee:
I am unclear on the "cargo fmt  -- --check" error.  This error doesn't occur on my local repository, and the CI error does not appear to give any additional insight as to the cause.

ceekdee:
@lulf:

- if it is too soon for another release of rust-lorawan, I can make the necessary adjustments using git rev or patch in embassy to pull in lorawan-device with this merged PR;
- the lorawan dependency in rust-lorawan appears to need to be changed to 0.7.2.  Should I include this in this PR?
- I do not understand the CI failure, as noted above.

lulf:
> @lulf:
> 
> * if it is too soon for another release of lorawan-device, I can make the necessary adjustments using git rev or patch in embassy to pull in lorawan-device with this merged PR;

I'm prepared to release it. Could you bump the minor version in the device/Cargo.toml?

> * the lorawan dependency in rust-lorawan appears to need to be changed to 0.7.2.  Should I include this in this PR?
> * I do not understand the CI failure, as noted above.
> 

Toolchain version is the only reason I can think it doesn't format the same way for you.
Make sure you have the latest stable or a recent nightly version, then run `cargo fmt` in the device/ .

> The lora-phy PR for embassy cannot be applied until this PR is merged into rust-lorawan.

ceekdee:
- Bumped device minor version to 0.10.0;
- Changed lorawan version in device/Cargo.toml to 0.7.2;
- Corrected fmt error.

lulf:
@ceekdee  Changes look good, looks like there is only 1 error left now :) It must be some differing toolchain version if you're not seeing them locally I think, I'm not aware of a way other than running the action on your own GH fork, i.e. if you submit a PR against a branch on your fork.

ceekdee:
@ivajloip  : fixing a clippy error in **encoding** requires your approval.  We'd appreciate your assistance.  Thanks!

ivajloip (approved):
It looks all good to me, thanks! :slightly_smiling_face:

PR #108, opened 2023-05-03
Update README.md, remove false hope for CN470

andelf:


lthiery:
Woops good call. I deprecated CN470 during the region overhaul. The new spec is confusing...

lthiery (approved):

andelf:
When will CN470 get supported? It is planned?
The design of `Region` enum forbids any customization. ☹️

lthiery:
> When will CN470 get supported? It is planned? The design of `Region` enum forbids any customization. frowning_face

What do you mean about "forbids customization"?

PR #109, opened 2023-05-05
Allow the use of builtin RNG for external PHY implementations

jbeaurivage:
During the integration efforts with the new `lora-phy` crate, a new `LoRaRadio` type was introduced. However, since it doesn't implement `RngCore` (where applicable), the `new_with_builtin_rng` constructor isn't useable. This PR aims to fix that.

## Notes on async
Now that `lorawan-device` optionally depends on `lora-phy`, we can do additional things wrt async RNGs. `lora-phy` is bound to provide a dedicated `AsyncRng` trait ([PR #9](https://github.com/embassy-rs/lora-phy/pull/9)) which we can use here. This PR includes provisions for using regular, `RngCore` based RNG implementations (either provided externally or included in the LoRa chip), or an `AsyncRng` implementation. The operating principle, in async mode, is to use a random numbers buffer that is pre-filled before it is needed. Use of an async implementation of the external LoRa PHY RNG can be enabled using the `async-rng` Cargo feature.

All operations in this crate use a bounded number of random numbers so that pre-filling to a known number is possible without having to re-fill the buffer mid-operation.


jbeaurivage:
This PR now includes provisions for async RNG implementations, see inital comment for additional details.

CBJamo:
I've used this PR with my hardware (rp2040 + custom sx1262) and have had no issues.

PR #110, opened 2023-05-05
Improve channel selection mechanism for join requests using fixed channel regional plans

### LINKED ISSUE #93: Use a specific subband when the network's subband is unknown
Is there a way to make a device use a specific subband when the network's (or gateway's) operating subband is unknown before joining? For example in the US915 region, if no subband is specified (`US915::new()`), then the device will randomly hop across the 64 available channels. 

Here is an interesting use case: when first attempting to join, the device tries an OTAA join request on random channels until one is accepted. Is there a way to then restrict the device to then only use the subband on which it successfully joined? Or at least manually change the region/subband of a created `Device`?

Forgive my ignorance; I'm pretty new to LoRaWAN. Is it common for networks to all operate on the same subband, or are gateways allowed to operate on different subbands?

jbeaurivage:
For US915 and AU915 regions, it would be beneficial to have a way of "biasing" the channels for join requests. This would enable quicker joins, instead of having to wait for a random channel selection to coincide with the network's enabled channels. This PR provides a way to achieve that.

Also fixes a bug in the datarate selection: If channel 63 is chosen, DR4 would've been selected. It is also very easy to implement the TODO to also pick from all 72 available channels:
```diff
- (random_number % 64) as u8
+ (random_number % 72) as u8
```

# Questions and concerns
* Currently the way to provide a "blessed" channel list is just by calling `join_biased` with an `&[u8]`. Obviously, this comes with its set of problems. If a list which is too long is provided, the function panics. Out of bounds channels are silently ignored.
* The channel list bias is also silently ignored for all dynamic channel plans.

Closes #93.


ivajloip (commented):
Hi Justin,

Thank you for sending this PR! This is really appreciated! Reading again issue #93 that seems good and I will let @lthiery and @lulf comment more as they are more familiar with that code than me.

What follows is a bit of a side comment, but I think it's still relevant. I completely agree that it's really nice to have a way to control the available channels when joining the network. My initial thought when seeing this PR was that for me in an ideal world, when we create a device we will set a few parameters like keys, region and possibly the initially enabled channels. Upon joining and through mac commands the enabled channels can be changed by the network and there is a config for that in the device, but this config could be set even before joining (so you can for example use a `default` method and a `with_channels`* one to create this config) and that will IMO provide a better interface than creating a new version of a few methods and it will still resolve the issue. Now, if other people think that such a restructuring can be useful, I (or they) can file an issue, we can discuss there the approximate format and work to move towards it. In any case I don't think such a discussion should be preventing this PR from being merged.

[*] We will even have an easy way to signal errors without panic as the `with_channels` method can return a `Result`.

jbeaurivage:
@ivajloip great! I agree that declaring the available channels before joining is probably better. However one thing I'd like to mention is that it's probably necessary to have a way of changing the available channels (or even enabling all channels) after the first join attempt has happened. That's because if the network is configured to use different channels than what is expected, it is still joinable - even if it'll take longer. My vision of how joining with a channel bias would work -
* Try joining with the expected channel bias for a certain number of times, say 10 tries
* If unsuccessful, make all the subsequent join attempts pick from all 72 channels at random

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -101,16 +101,33 @@ impl<const D: usize, F: FixedChannelRegion<D>> RegionHandler for FixedChannelPla
...
+                    Some(cl) => {
+                        let mut channel_list = heapless::Vec::<u8, 72>::new();
+                        for channel in cl {
+                            // Check that the specified channel is supported
+                            if *channel < 72 {
+                                channel_list.push(*channel).unwrap();
+                            }
+                        }
+                        // Pick one at random between all the supported channels.
+                        channel_list[random_number as usize % cl.len()]
```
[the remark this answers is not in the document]
jbeaurivage: `channel_list` contains only valid channels. The provided `u8` slice is looped on, and every channel is checked for validity (ie, < 72). If it's found to be valid, it's pushed into the `channel_list` `Vec`. Then, one of the valid channels in `channel_list` is picked at random. I suppose as it is now there are a few edge cases which could cause a panic:
* Specifying zero valid channels
* Specifying more than 72 valid channels (ie, contains redundant channels)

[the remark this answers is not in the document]
ivajloip: Sorry for not providing more details. The first condition seems partially correct. What I was thinking is what will happen if, for example,`cl = [42, 73]`. My reading is that in that case `channel_list == [42]` and if `random_number == 1` `random_number as usize % cl.len() == 1`. Then even though there are 72 values in the `channel_list` buffer, `channel_list[1]` will panic (I double checked it and for the latest version of heapless it prints `thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1'`). So I would suggest changing `random_number as usize % cl.len()` to `random_number as usize % channel_list.len()`, which should match your textual description.

ivajloip:
Yes, in my mind we can support the use case you provide and still have a better interface. For your case you initially create the device using a config created with `with_channels`. If that fails for say 10 tries, you create a new device config with the default channels and try again.

device/src/region/fixed_channel_plans/mod.rs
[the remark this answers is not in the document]
jbeaurivage: Oh damn you're right, my mistake! I'll change that.

jbeaurivage:
@ivajloip, I believe this is now closer that what you were envisioning. A new `Configuration` can be created with the `with_join_channels` method, which enables the user to selectively enable channels for the join process. The changes to the code are much simpler too, thankfully!

ivajloip:
Thank you for improving this! This looks really good to me! I would still like @lthiery or @lulf to also review this, but it looks good to me! :slightly_smiling_face:

ceekdee:
LoRaWAN® Regional Parameters RP002-1.0.4, section 2.5.2 details the preferred way to bias the search for a usable join channel - try one random channel in each sub-band first, then repeat using another channel in each sub-band.  Any channel bias parameterization assumes knowledge of the gateway capability topology in the area and may bias against a gateway that is capable of handling channels outside those known.

An overall problem here is leaving the join iterative process under user control. Various developers - including me - have been tripped up by a misunderstanding of the expected iterative nature of the join process for a fixed channel plan, and the reason for it.  It may be better to do the iterative process internally, only giving the user the ability to specify (for instance) the number of iterations allowed.

We are currently also trying to figure out the best approach for this in the LoRaWAN 1.0.4 implementation.

ceekdee:
I have been able to do enough testing with the LoRaWAN 1.0.4 implementation and my 8 channel TTN gateway to observe the following:

0.057861 INFO  deveui: ....
0.058074 INFO  credentials and configuration loaded from non volatile
0.058410 INFO  JOINING
5.636138 INFO  cf_list = **_ChannelMask([0, 255, 0, 0, 0, 0, 0, 0, 2])_**
5.661346 INFO  Network joined! ()

That channel mask shows that TTN wants me to use the second 125KHz sub-band and one 500KHz frequency for future send operations, which is what I would expect for my 8 channel gateway, which uses the second sub-band.  If we follow the standards-based  implementation (see previous comment) for attempting initial join, the second join attempt for my test environment would succeed and henceforth the channel bias would be established automatically.  In the worst case, it would take 9 attempts - one random channel from 9 sub-bands.  In the LoRaWAN 1.0.4 implementation, my current thought is to do the retries within the join function, only notifying the user if no connection can be made after attempting all active channels.

This would relieve some of the confusion users experience with join failure on only one attempt.  It's not a problem - retries are a normal occurrence when getting connected to an unknown gateway topology, especially for fixed channel plans like US915 with lots of channels to try.  Furthermore, it appears that channel bias will be established quickly for a 1.0.4 device without user intervention.

lthiery:
This is really nice work @jbeaurivage and you've done an excellent job explaining the logic in comments. However, I have two reservations about the approach:

1. I think it's elegant and simple to use a pre-programmed channel mask as you've done. However, it's clearly out of spec (I'll refer to the [1.0.4 regional spec](https://resources.lora-alliance.org/technical-specifications/rp002-1-0-4-regional-parameters) here) which states for US915 and AU915:
```
If using the over-the-air activation procedure, the end-device SHALL broadcast the Join-Request message alternatively on a random 125 kHz channel amongst the 64 channels defined using DR2 and on a 500 kHz channel amongst the 8 channels defined using DR6. The end-device SHOULD change channel for every transmission.
```

The spec does imply that it can take a while to join and instead recommends the following:
```
For rapid network acquisition in mixed gateway channel plan environments, the device SHOULD follow a random channel selection sequence that efficiently probes the octet groups of eight 125 kHz channels followed by probing one 500 kHz channel each pass.
```

Therefore, I believe the soundest way implement a join bias is to ensure that one's favorite subband is the first subband selected and then, after that, it cycles through subbands. The downside of staying within specification (as opposed to your approach) is that if the first Join fails on your favorite subband, you have to wait for the device to cycle all the way through before trying again.

2. If we don't agree with my point above, I would still urge that the stack internally clears the channel mask upon a few unsuccessful joins. I don't think it's a good idea to put the burden on the user to escape non-compliant join behavior. In fact, if the channel mask was cleared after a single unsuccessful Join, the behavior would be just as compliant as it was previous to this PR (there was no cycling of subbands originally). Actually, perhaps settings a default "reset channel bias after 1 join attempt" and letting the user override that to a larger number would fit all tastes. The default usage of this library would remain as compliant as it was previously, but for those that prefer faster joins on their subband at the expense of being non-compliant, they can explicitly increase this number.

jbeaurivage:
Thanks for the comments @ceekdee and @lthiery. I think your suggestions make a lot of sense. Whenever I find the time, I'll try to morph this PR into something that adheres more closely with the 1.0.4 spec. I do think it's a good idea to have the user choose their preferred subband without guaranteeing that it's the only subband that will be used, and giving that control back to the `lorawan-device` stack internally.

CBJamo:
I think the spec is short-sighted here. In private networks, it's not uncommon to use only one sub-band. In my use case, that's because there is only one gateway, and that gateway can only RX on one sub-band at a time. In that case it's pretty silly to wait for the system to cycle all the way around the bands, or even worse to random walk back the one sub-band that actually works. 

> Actually, perhaps settings a default "reset channel bias after 1 join attempt" and letting the user override that to a larger number would fit all tastes. The default usage of this library would remain as compliant as it was previously, but for those that prefer faster joins on their subband at the expense of being non-compliant, they can explicitly increase this number.

I like this recommendation, as in my case I could just set that number to whatever it's max is. If the device doesn't connect after u\<whatever\>::MAX attempts it's probably never going to connect anyway.

lthiery:
> I think the spec is short-sighted here. In private networks, it's not uncommon to use only one sub-band. In my use case, that's because there is only one gateway, and that gateway can only RX on one sub-band at a time. 

Well, there are a few things at play here. I wasn't around at the time but I think the fixed channel plans came out very early in the history of LoRaWAN and were mostly considering 64 channel gateways as the standard. I might be completely making this up, but I believe there's more 8-channel gateway deployments today than originally anticipated and dynamic channel plans are proving to provide better UX (join times) and flexibility for these; the flexibility where a single "standard N join channels" can still work across many national boundaries as the extra channels are tuned to specific legal or uncrowded frequency is great. Between that and rapid join times, I think there's a general movement to adopt more dynamic channel plans.

In your case, since you are just operating on your own private network, what if you deployed AS923? I believe there are plenty of AS923 configurations that are compliant with FCC.

Concerning this specific annoying characteristic, I've had to think through it before as I've seen the Helium Network implement band changes in Australia. Or in a more general use-case of "I have a US915/AU915 sensor, but it's not working! (because it's been constrained to a specific subband by the vendor or a previous user)". The lack of flexibility in hard-coding can create bigger problems than the join inconvenience in my opinion, where you might have to reprogram the sensor which might even require a trip out to the field.

Anyway, I totally agree that slow joins are terrible. I just hope the extra bit of context helps tolerate it better (as it has for me :laughing:).

jbeaurivage:
@ceekdee, @lthiery, I have updated this to match the spec more closely.

The new channel selection behavior on join attempts is now:

* By default, select one random channel per subband, including both bandwidths, in sequence. Ie, `[0, 7], [8, 15], ..., [64. 71]`.
* However, if the user chooses to construct a `Configuration` using `new_with_preferred_channels`, they get to pick which channels they would like to use for the first join attempts. They also have to specify a number of retries using those channels, after which the stack will revert to default  behavior. This method comes with a warning in the docs encouraging users to use a low number of retries, or else they run the risk of never being able to join the network.

device/src/region/mod.rs
```diff
@@ -175,6 +199,101 @@ impl Configuration {
...
+                match &mut config.state {
+                    State::US915(s) => {
+                        s.set_preferred_join_channels(preferred_channels, num_retries)
+                    }
+                    State::AU915(s) => {
+                        s.set_preferred_join_channels(preferred_channels, num_retries)
+                    }
+                    _ => (),
+                }
+                // let empty_mask = ChannelMask::<2>::new_from_raw(&[0x00, 0x00]);
```
lthiery, line 266: Looks like a lot of dead code here in comments.

device/src/region/mod.rs
```diff
@@ -175,6 +199,101 @@ impl Configuration {
...
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(Configuration)` if the provided channel set is correct and the
+    ///   region is a fixed channel plan
+    /// * The length of `channel_list` must be <= 16, otherwise `Err` will
+    ///   be returned.
+    /// * If a channel out of the specified channel range is specified,
+    ///   `Err` will be returned (ie, >= 72).
+    pub fn new_with_preferred_channels(
```
lthiery, line 240: Perhaps two constructors and the potentially out of spec one features this in the name?

For example, rename this function with `new_with_preferred_channels_and_noncompliant_retries` and then
```rust
new_with_preferred_channels(
        region: Region,
        preferred_channels: &[u8],
)  -> Result<Configuration, Error> {
   Self::new_with_preferred_channels_and_noncompliant_retries(region, preferred_channels, 1)
}
```

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -8,15 +8,71 @@ mod us915;
...
+    // List of subbands that haven't already been tried.
+    available_subbands: heapless::Vec<u8, 9>,
+}
+
+impl JoinChannels {
+    fn select_channel(&mut self, rng: &mut impl RngCore) -> u8 {
+        let random = rng.next_u32();
+
+        if let Some(fav) = &mut self.preferred_channels {
+            if fav.num_retries > 0 {
```
lthiery, line 31: It looks like this retry logic will only apply to the first Join. It's probably good to not mutate the `num_retries` after construction but set a retry counter for ever Join sequence.

device/src/region/mod.rs
```diff
@@ -175,6 +199,101 @@ impl Configuration {
...
+        region: Region,
+        preferred_channels: &[u8],
+        num_retries: usize,
+    ) -> Result<Configuration, Error> {
+        use Region::*;
+        match region {
+            US915 | AU915 => {
```
lthiery, line 247: The existence of this functions seems like a code-smell to me.

Now that I think about it, this logic might be scoped wrong and might belong in fixed_channel_plans/mod.rs instead? If that works, then you can avoid the `UnsupportedRegion` as well.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -8,15 +8,71 @@ mod us915;
 pub(crate) use au915::AU915;
 pub(crate) use us915::US915;
 
+#[derive(Clone)]
+struct PreferredJoinChannels {
+    channel_list: heapless::Vec<u8, 16>,
```
lthiery, line 13: It seems like your use of ChannelMask from the previous iteration has been abandoned in favor of this? Why may I only have 16 preferred channels?

I think that if you scope this back down to `fixed_channel_plans/mod.rs` as I mentioned elsewhere, it might be easy to implement use `ChannelMask<9>` here. You could create some constructors that assist in creating ChannelMasks that correspond to what users may typically do, such as `new_with_subband`.

One potential benefit of this approach too is that our stack could potentially "learn" a favorite subband upon Join and use that in future Joins (with default retries = 1). It's perhaps best to leave that feature for another PR, but I do think that speaking of "preferred subband" in terms of a "channel mask" has benefits in that regard as well.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -8,15 +8,71 @@ mod us915;
...
+    // Decrementing number representing how many tries we have left with the specified list before
+    // reverting to the default channel selection behavior.
+    num_retries: usize,
+}
+
+#[derive(Default, Clone)]
+struct JoinChannels {
+    preferred_channels: Option<PreferredJoinChannels>,
+    // List of subbands that haven't already been tried.
+    available_subbands: heapless::Vec<u8, 9>,
```
lthiery, line 23: Could this be implemented with bool flags? Start with all bits high and but them low as you try each subband.

lthiery (changes requested):
Thanks for taking another pass at this @jbeaurivage! 

I have a few suggestions, some stronger than others but let me know if you disagree with any - I'm happy to discuss further. I'm on vacation until next week but wanted to give you some quick feedback and I can try running and testing the code next week.

device/src/region/mod.rs
jbeaurivage: The problem is that whether the actual function lives in `region/mod.rs` or `fixed_channel_plans/mod.rs`, at the lowest level the end user can only construct a `Region`. That enum makes no distinction between fixed and dynamic channel plans. There would probably be a way to type-encode that, but it would probably involve creating more intermediary types, and I'm not sure the added complexity is so worth it.

device/src/region/fixed_channel_plans/mod.rs
jbeaurivage: I made a 16 channel limit just in the spirit of saving memory. There's no technical reason, but I figured I could easily change it back to 72 if needed. It's just that usually gateways will have a max of 16 channels enabled (actually 18 counting 500 kHz channels).

One reservation I have about this is I don't find the `ChannelMask` API super intuitive. It's more oriented towards channel banks rather than individual channels. I definitely understand why it's designed that way, but it makes it harder to use in non-MAC contexts.

 One of my goals with this PR is to keep the number of RNG calls bounded, in preparation for if #109 gets accepted - if that is a feature people want, obviously. That's done much more easily if we hold an array of enabled channels, instead of looping randomly through all channels until we find one that is enabled. 
However now that I think of it, maybe temporarily collecting enabled channels into a `Vec` when needed would do the trick.

device/src/region/mod.rs
lthiery: I think we could look at the [`with_state`](https://github.com/ivajloip/rust-lorawan/blob/master/device/src/region/mod.rs#L178) function. Either use it directly or do something in that theme.

Essentially, the user could construct US915/AU915 "state" and do the join bias and then build a configuration there. Really, we should probably just implement From<Region> for Configuration for all "states", then the user can make tweaks and "into" it when constructing a device.

device/src/region/fixed_channel_plans/mod.rs
lthiery: > I made a 16 channel limit just in the spirit of saving memory.

`heapless::Vec<u8, 16>` is at least 16 bytes. `ChannelMask<9>` is 9 bytes.

> One reservation I have about this is I don't find the ChannelMask API super intuitive. 

I agree, but that's what I recommend we extend and work on rather than creating a novel type which serves the same function. Two types that functionally do the same thing doesn't sit well with me, personally.

device/src/region/mod.rs
lthiery: Sorry if my last thought there is not clear - I have limited time today. I have a flight tmrw and may try to draft what I mean and PR to your branch.

Generally though, I think it’s quite possible to make these errors compile-time impossible without much complexity. I think that’s always better design than throwing errors are runtime.

lthiery: Threw up this PR to show what I mean: https://github.com/jbeaurivage/rust-lorawan/pull/1

device/src/region/fixed_channel_plans/mod.rs
lthiery: > One of my goals with this PR is to keep the number of RNG calls bounded. [...] That's done much more easily if we hold an array of enabled channels, instead of looping randomly through all channels until we find one that is enabled.

Your solution to minimizing RNG calls creates additional RAM overhead as one has to store a channel list for indexing instead of using the ChannelMask. I would isolate that kind of change to its own PR since that goal adopts an alternate design trade-off than what is currently adopted in this repository. 

Up until now, my understanding is that we should prioritize minimizing RAM usage, since RAM is generally the most precious resource on micro-controllers (as opposed to using a few more instructions to generate additional entropy).

lthiery: @jbeaurivage [I've added a PR](https://github.com/ivajloip/rust-lorawan/pull/114) that I think will completes the function set needed for manipulating ChannelMask.

I believe an API like this would be a good start:

```rust 
enum Subband {
   _1,
   ...
  _8,
}
impl<const D: usize, F: FixedChannelRegion<D>> FixedChannelPlan<D, F> {
    fn set_preferred_join_subband(subband: Subband) {
         // here you enable the appopriate bank of channels and then enable the appropriate fat channel using the new ChannelMask API
    }
}
```

This would cover 95% of these "preferred join channel" use-cases as networks tend to standardize on a specific subband. The use of the enum would maintain a safe public API. You could still provide a more dangerous `set_preferred_join_channel` API off of the FixedChannelPlan trait for special cases where users might have different needs :shrug:

device/src/region/fixed_channel_plans/mod.rs
jbeaurivage: What constitutes a join sequence? As far as I can tell, the user is responsible for calling `join()` until the network replies.

device/src/region/mod.rs
jbeaurivage: @lthiery, take a look at the latest revision. I've managed to make specifying preferred channels completely statically checked. Although, now the user has to provide a reference to a fixed-size array rather than a slice. I think the benefits outweigh the hit on ergonomics, but I'd definitely like your opinion on the matter.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -2,21 +2,170 @@ use super::*;
...
+);
+
+impl From<Channel> for u8 {
+    fn from(value: Channel) -> Self {
+        value as u8
+    }
+}
+
+#[derive(Clone)]
+struct PreferredJoinChannels {
```
lthiery, line 42: You'll notice if you build with all features enabled, we have an issue since this struct doesn't implement Serialize/Deserialize. If you keep digging in, you'll see heapless::Vec doesn't implement Serialize/Deserialize.... so we'd have to figure that out if we maintain this heapless::Vec of preferred channels. We could of course write a Serialize/Deserialize implementation...

However, I'm curious what this looks like if "preferred channels" are described in a `ChannelMask<9>`. We already have functions to enable/disable channels and subbands on that struct. And we know how to derive valid channels from that struct and after `max_num_retries`, we could reset the channel mask to default (ie: all enabled).

And just to be clear, I'm not saying that that approach is "the way". I just have a feeling things will look simpler with it. I'll maybe try to PR against your branch so we can see how it looks comparatively.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -2,21 +2,170 @@ use super::*;
...
+            return p;
+        }
+
+        if self.available_subbands.is_empty() {
+            self.available_subbands.reset();
+        }
+
+        // Unwrapping is ok because it should never be empty by this point
+        let subband = self.available_subbands.pop_next().unwrap();
+        8 * subband + (rng.next_u32() % 8) as u8
```
lthiery, line 128: this means we will only be joining on the lower 64 channels if I'm reading this right?

device/Cargo.toml
```diff
@@ -21,11 +21,13 @@ rustdoc-args = ["--cfg", "async"]
 lora-modulation = { version = ">=0.1.2", default-features = false }
 lorawan = { version = "0.7.4", path = "../encoding", default-features = false }
 heapless = "0.7"
+bitvec = { version = "1.0.1", default-features = false, features = ["atomic"] }
```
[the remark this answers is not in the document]
jbeaurivage: Right, that was an unfinished idea. I'll remove it next time I touch this.

device/src/region/fixed_channel_plans/mod.rs
jbeaurivage: Yep. I'll see what it looks like. Take all of the latest changes with a grain of salt, the PR isn't finished. I just wanted your input on the compile time checking side of things before I kept going.

lthiery:
As I'm testing out the ChannelMask<9> idea, I'm also noticing that there might be no way for the user to actually construct a device with the join bias?

In [my previous PR]([device/src/region/mod.rs](https://github.com/jbeaurivage/rust-lorawan/pull/1/files#diff-e69ecdc6fbc006343caf6254974dbec55238bb26d62a34ec0cf9a1085910e7af)), I did some wiggling of nested structs to make it so the user could have access to specific APIs:
```
device/src/region/mod.rs
device/src/region/fixed_channel_plans/us915/mod.rs
device/src/region/fixed_channel_plans/au915/mod.rs
```

I can try to reimplement the idea in a separate PR than the ChannelMask one I'm playing with atm.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -2,18 +2,59 @@ use super::*;
...
+mod join_channels;
+use join_channels::JoinChannels;
+
 mod au915;
 mod us915;
 
 pub(crate) use au915::AU915;
 pub(crate) use us915::US915;
 
+seq_macro::seq!(
```
lthiery, line 14: I think we can delete this channel now actually since we only expose Subbands

lthiery (approved):

PR #111, opened 2023-05-13
Add repo-wide rules for the code formatting

ivajloip:
I have noticed a few PR that wrapped comments back and forth recently and in an attempt to stop this, I propose some formatting rules. I am looking forward to discussing them and agreeing on a set of rules that the tool will always apply. I ran the formatting and I have included the relevant result. I am not 100% happy with it, but I can live with it.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,23 @@
+# This file defines the Rust style for automatic reformatting.
+# See also https://rust-lang.github.io/rustfmt
+
+# Rust language edition.
+edition = "2021"
+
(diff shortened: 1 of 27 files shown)


lucasgranberg:
Good idea. I don't care what the rules are as long as there are some and my IDE can format as I type.

CBJamo:
Strong agree. So long as rustfmt understands the rules and automatically applies them it should be good to go. Then we just need to require that PRs have rustfmt run against them.

lulf (approved):

PR #112, opened 2023-06-25
Add derives to make Lora Configuration (Region, JoinMode) serializable and testable

ShakenCodes:
While trying to store storing LoRaWAN configuration data in Flash we tried to serialize the Region and JoinMode, but the latter has not derived the serde traits. This PR adds derived that allow serde use for JoinMode, and make Region and JoinMode copyable for moving and testing purposes, and comparable for unit test purposes.


lulf (approved):
Thanks!

PR #114, opened 2023-08-06
add function to ChannelMask for setting or clearing individual channels

lthiery:
This provides a function for enabling or disabling a specific channel.


encoding/src/maccommands.rs
```diff
@@ -521,6 +521,23 @@ impl<const N: usize> ChannelMask<N> {
         self.0[index] = value;
     }
 
+
+    /// Enable or disable a specific channel. Recall that LoRaWAN channel numbers start indexing
+    /// at zero.
+    ///
+    /// Improper use of this method could lead to out of bounds panic during runtime!
+    pub fn set_channel(&mut self, channel: usize, set: bool) {
```
ivajloip, line 529: Did you consider creating a builder for this type? I can be convinced that this case is too simple and having a builder is really an overkill, 🙂

encoding/src/maccommands.rs
```diff
@@ -521,6 +521,23 @@ impl<const N: usize> ChannelMask<N> {
...
+        let index = channel / 8;
```
ivajloip, line 530: nit(optional): for consistency with`channel_enabled`, I would prefer to have bit operations here.

encoding/tests/maccommands.rs
```diff
@@ -328,6 +328,18 @@ fn test_channel_mask() {
...
 
+#[test]
+fn test_channel_mask_enable_and_disable_channel() {
+    let data = vec![0x00];
+    let mut chan_mask = ChannelMask::<1>::new(&data[..]).unwrap();
+    chan_mask.set_channel(7, true);
+    let expected = vec![false,false,false,false,false,false,false,true,];
+    assert_eq!(&chan_mask.statuses::<8>()[..], &expected[..]);
+    chan_mask.set_channel(7, false);
+    let expected = vec![false,false,false,false,false,false,false,false,];
```
ivajloip, line 339: nit: You could use something like `[false; 8]` here.

encoding/tests/maccommands.rs
```diff
@@ -328,6 +328,18 @@ fn test_channel_mask() {
...
+    assert_eq!(&chan_mask.statuses::<8>()[..], &expected[..]);
+}
```
ivajloip, line 341: Maybe it would be better to have a test with byte index > 0 (probably using something like `let expected = [false; 16]; expected[11] = true;` for example).

ivajloip (approved):

encoding/src/maccommands.rs
lthiery: Good point! Fixed

encoding/tests/maccommands.rs
lthiery: That is better. Fixed

encoding/tests/maccommands.rs
lthiery: Good point! Thanks

ivajloip (approved):

lthiery:
> And thank you for proposing this change and agreeing to accommodate my recommendations! 🙂

No problem! I appreciate the review :)

Issue #117, opened 2023-09-04
async_device RX windows may be disturbed by noise


lthiery:
There's a bit of a problem in the call logic where the timing and windows are managed in [`rx_with_timeout`](https://github.com/ivajloip/rust-lorawan/blob/7f7ceef024991d0f068b59cff715897b021e8521/device/src/async_device/mod.rs#L451-L523). _Any_ received radio data on the proper channel at the proper time will essentially lead us to executing [lorawan_parse](https://github.com/ivajloip/rust-lorawan/blob/7f7ceef024991d0f068b59cff715897b021e8521/device/src/async_device/mod.rs#L311) on the data. The problem lies in the fact that if it's unsuccessful, we don't continue to listen at all and the whole listening and receive context is long gone.

PR #118, opened 2023-09-04
refactor mac for state machine implementation

lthiery:
This is the first stage in some refactoring that I am doing to try to unify the async and no_async (state_machine) device implementations. 

I've minimized impact on the async_device area of the code to try to keep things simple. However, I believe that this should allow us to move more non-runtime logic into the `mac` module and to share more of the protocol logic amongst the two runtimes.

I ran into this as I was working on a solution for #117 and I realized it would be a lot easier to start consolidating some of the duplicate code.

EDIT: to be clear, I think this refactor starts to make it easier to share the [logic here](https://github.com/ivajloip/rust-lorawan/blob/7f7ceef024991d0f068b59cff715897b021e8521/device/src/async_device/mod.rs#L311-L360).

### DIFF
--- a/device/src/lib.rs
+++ b/device/src/lib.rs
@@ -2,8 +2,6 @@
 #![cfg_attr(feature = "async", feature(async_fn_in_trait))]
 #![allow(incomplete_features)]
 
-extern crate alloc;
-
 use heapless::Vec;
 
 pub mod radio;
--- a/device/src/lib.rs
+++ b/device/src/lib.rs
@@ -2,15 +2,15 @@
 #![cfg_attr(feature = "async", feature(async_fn_in_trait))]
 #![allow(incomplete_features)]
 
+extern crate alloc;
+
 use heapless::Vec;
 
 pub mod radio;
 
 mod mac;
+pub use mac::types::*;
 use mac::Mac;
-
-mod types;
-pub use types::*;
 
 pub mod region;
 pub use region::Region;
@@ -22,7 +22,7 @@ use lorawan::{
     parser::{DecryptedDataPayload, DevAddr},
 };
 use state_machines::Shared;
-pub use state_machines::{no_session, no_session::SessionData, session};
+pub use state_machines::{no_session, session};
 
 pub use rand_core::RngCore;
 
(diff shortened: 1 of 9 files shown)


device/src/lib.rs
```diff
@@ -2,16 +2,16 @@
 #![cfg_attr(feature = "async", feature(async_fn_in_trait))]
 #![allow(incomplete_features)]
 
+extern crate alloc;
```
lulf, line 5: This should be behind a feature flag for the cases where alloc is not supported (most embassy examples are no alloc), but I also can't see any place in the code that requires it?

device/src/mac/mod.rs
```diff
@@ -0,0 +1,178 @@
+use crate::{
+    radio::RadioBuffer, region, state_machines::Downlink, FcntDown, FcntUp, SendData, SessionKeys,
+};
+use generic_array::{typenum::U256, GenericArray};
```
lulf, line 4: Just a note: it would be good to change this to use const generics instead of generic array at some point.

lthiery: Good point. Left an issue to track it #126

device/src/lib.rs
lthiery: Woah! Thanks for catching that. It's really not meant to be there. My editor does that sometimes and now I'm motivated to disable that feature.

lthiery:
FYI, I'm also considering getting some tests setup around this one. It should be quite a bit easier than the async ones.

lulf (dismissed):

device/src/mac/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+                        Some(Response::SessionExpired)
+                    } else {
+                        Some(Response::DownlinkReceived(fcnt))
+                    };
+                }
+            }
+        }
+        None
+    }
+    pub(crate) fn rx2_elapsed(&mut self) -> Response {
```
ivajloip, line 106: nit: adding a new line above can improve readability.

ivajloip (approved):
I am getting tired - long day, but I tried to quickly scan this and it looks good, thanks! :)

lulf (approved):

PR #119, opened 2023-09-04
async_device: rx config decoupling & sleep

lthiery:
I was considering solutions to #117 but I wanted to do a first "simple" PR to try to make it easier.

First, I decouple configuring the radio to receive with actually waiting for the radio to receive. This will make it easier to configure the radio once and receive many packets (this will also be important for Class C behavior).

Since I was in this area of the code, I've "bundled" the logic with adding some sleeps. This will also help with class C behavior as instead of "sleeping", an RXC window will be opened.

### DIFF
--- a/device/src/async_device/lora_radio.rs
+++ b/device/src/async_device/lora_radio.rs
@@ -21,7 +21,7 @@ where
     DLY: DelayUs,
 {
     pub fn new(lora: LoRa<RK, DLY>) -> Self {
-        Self { lora, rx_pkt_params: None,}
+        Self { lora, rx_pkt_params: None }
     }
 }
 
--- a/device/src/async_device/lora_radio.rs
+++ b/device/src/async_device/lora_radio.rs
@@ -13,6 +13,7 @@ where
     DLY: DelayUs,
 {
     pub(crate) lora: LoRa<RK, DLY>,
+    rx_pkt_params: Option<lora_phy::mod_params::PacketParams>,
 }
 impl<RK, DLY> LoRaRadio<RK, DLY>
 where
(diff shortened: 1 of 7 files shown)


lthiery:
CI was failing because of the lora-phy feature. I'm not sure why we have a downstream trait implementation here? 

https://github.com/ceekdee/lora-phy/pull/1

lthiery:
I ended up just updating the dependency to the new API, so it's re-included.

lthiery:
@ceekdee and/or @lucasgranberg could you review the API implementation that I did for lora-phy?

lulf (approved):
Is this tested on real hardware? Changes to this part of the code has turned out to easily break things in the past. Otherwise looks good.

lthiery:
> Is this tested on real hardware? Changes to this part of the code has turned out to easily break things in the past. Otherwise looks good.

I'm still trying to get my test hardware running. I was thinking of calling this stack in [my simulator](https://github.com/helium/virtual-lorawan-device) for starters though.

lthiery:
Did a bit of clean-up on those commits. I'll get this tested ASAP

lthiery:
Actually, I think I can rig up a few unit tests that exercise this code here with tokio... that might be best long-term

lulf (dismissed):
This looks good 🚀

lthiery:
I rebased this PR which removed a lot of noise around the lora-phy device. Ready to go once we can land #125

ivajloip (approved):
IIUC this is built on top of [PR 125](https://github.com/ivajloip/rust-lorawan/pull/125) and as such the same comments apply. Otherwise it looks good to me.

lthiery:
I pushed the formatting recommendations from #125 into here and rebased on master.

lulf (approved):

PR #120, opened 2023-09-04
add async and external-lora-phy to stable CI

lthiery:


lthiery:
oh derp. still need async fn in trait

PR #123, opened 2023-09-04
standalone lora-modulation crate

lthiery:


lulf (commented):
Looks good, but I'm wondering if this crate should ideally be placed in an organisation and a repository that can enjoy wider access to maintainers and reviewers? Also I think 'lora' is a too generic name, and something like 'lora-types' or 'lora-modulation' would be more appropriate.

lthiery:
@ceekdee added you as a code maintainer to this module as well as added your bandwidth comment from #123

lthiery:
@ivajloip we can give @ceekdee write access? i think the way you have it setup, it lets him push branches and review PRs within his code ownership purview?

ivajloip (approved):
I hope that's fine now.

ivajloip:
and to answer @lulf, I think it could make sense to have an organization. Let's chat somewhere about how that could work (what's the ideal end state, how to get there, etc).

PR #124, opened 2023-09-08
implement From<SF/BW> for u32

lthiery:
When trying to integrate with the lora-phy crate, I realized these functions were missing.


lthiery:
@ivajloip looks like i can't request @ceekdee as a reviewer yet, so something is missing :thinking:

lthiery:
Supports https://github.com/embassy-rs/lora-phy/pull/26

lthiery:
Follow up on @ceekdee comment above - it seems he did not accept the contributor role so that explains that.

A few comments about other changes here:
* i realized the other directories omit the "lora" or "lorawan" part. I renamed to modulation to fit that theme.
* when working on the testing PR, I realized that "defmt" and "serde" needed to transitively activate the features in "lora" and "lorawan (encoding)" respectively.

lulf (approved):
LGTM!

PR #125, opened 2023-09-10
Async Tests

lthiery:
I've put together some tests around the async runtime so that we may proceed with #119 with more confidence.

The tests:
* demonstrate RX1 and RX2 joining (which more or less demonstrates that RX1 and RX2 works for data due to how @lulf architected the code)
* RX timeout when no RF is received
* the issue from #119 where a single noisy frame breaks things.

While implementing JoinAccept for the tests, I inadvertently uncovered a bug with the JoinAcceptCreator which made it not tolerate building a JoinAccept without a CFList. I implemented a small patch to the issue which makes it tolerate empty CFList or full Type 1 CFList. I think it still needs to tolerate partial Type 1 CFList and I'm unclear whether it tolerates Type 2 CFList packing, but that seemed out of scope for now and can be addressed in a future PR.

### DIFF
--- a/device/src/async_device/test/mod.rs
+++ b/device/src/async_device/test/mod.rs
@@ -1,0 +1,142 @@
+use super::*;
+use crate::radio::{RfConfig, RxQuality, TxConfig};
+use crate::region;
+use lorawan::default_crypto::DefaultFactory;
+
+mod timer;
+use timer::{TestTimer, TimerChannel};
+
+mod radio;
+use radio::{RadioChannel, TestRadio, Uplink};
+
+type Device =
+    crate::async_device::Device<TestRadio, DefaultFactory, TestTimer, rand_core::OsRng, 512>;
+
+pub fn get_key() -> [u8; 16] {
+    [0; 16]
+}
+pub fn get_credentials() -> JoinMode {
+    JoinMode::OTAA { deveui: [0; 8], appeui: [0; 8], appkey: get_key() }
+}
+
+fn setup() -> (RadioChannel, TimerChannel, Device) {
+    let (radio_channel, mock_radio) = TestRadio::new();
+    let (timer_channel, mock_timer) = TestTimer::new();
+    let region = region::US915::default();
+    let async_device: crate::async_device::Device<
+        TestRadio,
+        DefaultFactory,
+        TestTimer,
+        rand_core::OsRng,
+        512,
+    > = Device::new(region.into(), mock_radio, mock_timer, rand::rngs::OsRng);
+    (radio_channel, timer_channel, async_device)
+}
+
+#[tokio::test]
+async fn test_join_rx1() {
+    let (radio, timer, mut async_device) = setup();
+    // Run the device
+    let async_device = tokio::spawn(async move { async_device.join(&get_credentials()).await });
+
+    // Trigger beginning of RX1
+    timer.fire().await;
+    // Trigger handling of JoinAccept
+    radio.handle_rxtx(handle_join_request);
+
+    // Await the device to return and verify state
+    if let Ok(()) = async_device.await.unwrap() {
+        // NB: timer is armed twice (even if not fired twice)
+        // because RX1 end is armed when packet is received
+        assert_eq!(2, timer.get_armed_count().await);
+    } else {
+        assert!(false);
+    }
+}
+
+#[tokio::test]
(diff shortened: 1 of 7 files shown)


lthiery:
I've refactored the test fixtures to make the unit tests a bit easier to read & scale.

Now the mock radio and timer are basically channels awaiting messages from the text fixtures (instead of being jerry-rigged with flags and state upfront). This makes it easier to "step" through the async states in the unit test and, for example, to see what downlink has been configured for that particular unit test.

lulf (approved):
Thanks for getting the tests there!

lthiery:
Great! I'd like to land this first and then proceed with #119 if possible. Need a +1 from @ivajloip though because of changes made to the JoinAcceptCreator.

lulf (approved):

device/src/async_device/test/timer.rs
```diff
@@ -0,0 +1,49 @@
+use crate::async_device::radio::Timer;
+use std::sync::Arc;
+use tokio::sync::{mpsc, Mutex};
+impl TestTimer {
```
ivajloip, line 4: nit: Please add a new line above to improve readability.

device/src/async_device/test/radio.rs
```diff
@@ -0,0 +1,96 @@
+use super::*;
+use crate::async_device::radio::PhyRxTx;
+use std::sync::Arc;
+use std::vec::Vec;
+use tokio::{
+    sync::{mpsc, Mutex},
+    time,
+};
+type RxTxHandler = fn(Option<Uplink>, RfConfig, &mut [u8]) -> usize;
```
ivajloip, line 9: nit: please add a line above to improve readability.

device/src/async_device/test/radio.rs
```diff
@@ -0,0 +1,96 @@
...
+impl TestRadio {
```
ivajloip, line 10: the same here

device/src/async_device/test/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    timer.fire().await;
+    // Trigger end of RX1
+    timer.fire().await;
+    // Trigger start of RX2
+    timer.fire().await;
+    // Pass the join request handler
+    radio.handle_rxtx(handle_join_request);
+
+    // Await the device to return and verify state
+    if let Ok(()) = async_device.await.unwrap() {
```
ivajloip, line 73: I might be missing something, but it seems to me that using `is_ok` produces a more readable code here.

device/src/async_device/test/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    timer.fire().await;
+    // Trigger end of RX1
+    timer.fire().await;
+    // Trigger start of RX2
+    timer.fire().await;
+    // Trigger end of RX2
+    timer.fire().await;
+
+    // Await the device to return and verify state
+    if let Err(Error::RxTimeout) = async_device.await.unwrap() {
```
ivajloip, line 96: similarly with `is_err` here.

device/src/async_device/test/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    let (radio, timer, mut async_device) = setup();
+    // Run the device
+    let async_device = tokio::spawn(async move { async_device.join(&get_credentials()).await });
+    // Trigger beginning of RX1
+    timer.fire().await;
+    // Send an invalid lorawan frame. 0 length is enough to confuse it
+    radio.handle_rxtx(|_, _, _| 0);
+
+    // Await the device to return and verify state
+    if let Err(Error::UnableToDecodePayload(_)) = async_device.await.unwrap() {
```
ivajloip, line 114: and here.

ivajloip (approved):
Thanks for adding those tests! I think they are really useful. I tried to add a few quick comments that will hopefully improve the code a bit.

device/src/async_device/test/mod.rs
lthiery: Agree with the previous ones, but in this case, I am matching on a specific `Error::RxTimeout`

device/src/async_device/test/mod.rs
lthiery: Same here: I am matching on a specific error: UnableToDecodePayload

PR #131, opened 2023-09-14
update lora-phy to v2 and remove external-lora-phy from modulation

lthiery:
The async-device implementation with lora-phy is updated to 2.1.1.

`lora-phy` now depends on the local repo crate `lora-modulation` v0.1.2, therefore we can remove feature from `lora-modulation`.

The rust-toolchain file is also updated to match the lora-phy nightly which matches embassy's nightly.

Also, a little bit of clippy is cleaned up.


lthiery:
DNM yet. Found some things to fix

lthiery:
All is resolved, I believe.

lulf (approved):

PR #133, opened 2023-09-16
set all crates to rust edition 2021. set workspace resolver to "2"

lthiery:
I was getting a compilation warning:
```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

Not too sure what the risks are but I figured I just set everything to most recent. At any rate, the warning is resolved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "encoding",
     "modulation",
 ]
+resolver = "2"
 
 [profile.dev]
 opt-level = "s"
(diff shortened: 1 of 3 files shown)


ivajloip (dismissed):
Would editing all the editions alone fix the issue. Otherwise I think it's fine also the way the PR is (but I am also not sure about all the implications).

lthiery:
I tried to skip the resolver piece but for some reason it didn't work. Maybe I held something wrong though :shrug:

lulf (approved):

lulf (approved):

PR #135, opened 2023-09-17
Expand on device unit tests

lthiery:
Some pieces from the `async_device` unit tests are refactored into the `lorawan_device::test_utils` module.

The `state_machine` device is tested almost completely now and a few bugs were even found, such as

```rust
Response::NoAck => crate::Response::NoUpdate,
```
should've been 
```rust
Response::NoAck => crate::Response::NoAck
```

And WaitingForRx was not transitioning `into_idle` after receiving the downlink successfully.

Some small changes were made to encoding, but should be non-substantive (just trying to improve readability). Plus some simple type conversions.

### DIFF
--- a/device/Cargo.toml
+++ b/device/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "lorawan-device"
 version = "0.10.0"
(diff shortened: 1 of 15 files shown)


lulf (approved):
🚀

Issue #134, opened 2023-09-17
Use specific DevEui, AppEui, AppKey, NewSKey, AppSKey, DevAddr Types in Encoding


lthiery:
It's currently inconsistent in its use of AES128 type:
```rust
pub enum JoinMode {
    OTAA { deveui: [u8; 8], appeui: [u8; 8], appkey: [u8; 16] },
    ABP { newskey: AES128, appskey: AES128, devaddr: DevAddr<[u8; 4]> },
}
```

It's generally preferable to have specialized types, so OTAA should adopt AES128 as well. 

We probably want to make special types for DevEui and AppEui as well?

ivajloip:
Is [EUI64](https://github.com/ivajloip/rust-lorawan/blob/master/encoding/src/parser.rs#L912) something that will help here? Otherwise, sure, we can have separate types for those.

lthiery:
Yes, I do think EUI64 is appropriate for AppEui & DevEui. But we still might want to use a new type tho and do something like `struct AppEui(EUI64)`.

ivajloip:
sounds good! 🙂

lthiery:
#137 updates the device types to be more specific. It would still be good to update the encoding API to use these specific types as well. Work is being deferred for now as it should be paired with other breaking changes in the encoding library.

lthiery:
@ivajloip let me know how #177 looks to you

PR #137, opened 2023-09-18
more specific types for keys and euis

lthiery:
Addresses #134 

I did this while trying to minimize impact on encoding. If we like the approach, I believe these types should actually exist over in encoding.

### DIFF
--- a/device/src/async_device/mod.rs
+++ b/device/src/async_device/mod.rs
@@ -14,7 +14,7 @@ use heapless::Vec;
 use lorawan::{
     self,
     creator::DataPayloadCreator,
-    keys::{CryptoFactory, AES128},
+    keys::CryptoFactory,
     maccommands::SerializableMacCommand,
     parser::DevAddr,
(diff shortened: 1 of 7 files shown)


lthiery:
Also, at second glance, I think we probably just add AppEui and DevEui to replace EUI64. I don't think a new type around a single EUI64 adds anything.

lulf (approved):

ivajloip (approved):
Seems good and it's nice to see that things get more consistent now. Is the idea of renaming EUI64 to Eui to allow for EUIs that are not 64 bits? I am just wondering about the value of this change compared to potentially breaking things for other users.

lthiery:
@ivajloip - I agree, the rename as is is not worth a breaking change. I reverted that.

Splitting EUI64 into AppEui/DevEui depending on context though seems more useful though and I'll PR that idea separately so it can be discussed on its own. It will change the API in quite a few places though so it would make sense to try to get the `Error enum` landed around the same time, IMO.

lulf (approved):

PR #138, opened 2023-09-18
rename state_machines to nb_device

lthiery:
I think this rename expresses the reason why "state_machines" exists.


lulf (approved):
The terms usually used to distinguish these different behaviors in the rest of the rust ecosystem is `nb` (nonblocking), `blocking` or `async`. In that sense, the no_async_device could be nb_device. 

In any case I think there is some potential for consolidation. 

But the above is just my opinion, I think the PR looks good.

lthiery:
And yes I think there’s room for consolidation. I started stubbing out the Mac module as I think more lorawan stuff can be shared there (and shaped in a way to be compatible with the 1.0.4 fork). That will hopefully leave the async and no_async modules focused on some kind of runtime coordination logic… not quite sure how to describe it but maybe that makes sense.

lthiery:
@lulf i think i finally understood your `nb_device` suggestion and I think that's the right name! Update the PR here

ivajloip (approved):
Looks good to me!

PR #142, opened 2023-10-08
implemented proper RX1+RX2 windows, waiting for the preamble and receiving the rest of the packet regardless of the window timeout

ilya-epifanov:


ilya-epifanov:
Depends on https://github.com/embassy-rs/lora-phy/pull/33

device/src/async_device/mod.rs
```diff
@@ -454,105 +457,120 @@ where
...
+                Err(Error::RxTimeout)
+            }
+        }
+    }
+
     async fn rx_with_timeout_inner(
         &mut self,
         frame: &Frame,
         window_delay: u32,
+        expected_payload_length: usize,
```
lucasgranberg, line 530: How do you know how many bytes the network will send?

lthiery:
@ilya-epifanov Thanks for this! It looks interesting.

Just to clarify, the point of this PR is to ensure that we do not cancel a receive if the preamble has been detected? Can you confirm the high-level strategy:

Step one
```
select
    * preamble IRQ
    * window timeout
 ```

If the preamble IRQ fires first, then we do something like
```
select
   * RxDone
   * max potential payload timeout
```

I just want to confirm the high-level pieces before I attempt to drill into any details.

**Comment by ilya-epifanov** [comment]
@lthiery yep, that's exactly it.
It might be a good idea to expose the payload timeout (as in expected payload size) in the API

**Comment by lthiery** [comment]
> @lthiery yep, that's exactly it. It might be a good idea to expose the payload timeout (as in expected payload size) in the API

My initial impression is as follows: 
* port your `time_on_air_us` to the `modulation` crate with something like this as function signature: `fn calculate_toa_us(len: usize, sf: SpreadingFactor, bw: Bandwidth, cr: Option<CodingRate>) -> u32`. This is a generally very useful function and those working on p2p protocols may enjoy its utility if you place it there.
* I think you can continue doing the rest the way you are doing... it's just that technically, for data frames, you could ask the region::Configuration for max_rx_len and, given the coding rate, [it should be able to give that to you](https://lora-developers.semtech.com/documentation/tech-papers-and-guides/the-book/packet-size-considerations/). It's just, that functionality doesn't exist yet... You could try to implement it, but if you want to just get on with it, a max possible size seems fine, although 230 might be a bit short. I think both US915 and EU868 both max out around 222 for the **application payload** (ie: a Data FRMPayload). That's not accounting for the MHDR+FHDR (and FOpts) + MIC, I believe.  Glancing at the spec, I think that adds 1 + (4+1+2+15) + 1 =  24 bytes... so we're at 246 bytes total?

I need to spend more time thinking about your changes to the `async_device::PhyRxTx` trait. You've "polluted" the `rx_until_state` signature with the `DesiredIrqState` from `lora_phy`. In addition, your default implementations appear to lean on some structs from lora_phy as well. The trait (and it's default impls) needs to remain generic so we need to find another way around that, but I don't really have a solution to propose at the moment.

File: device/src/async_device/mod.rs (hunk 1 above)
Comment (lucasgranberg) [reply device/src/async_device/mod.rs:537]: ah I see, maybe call it something with maximum

File: device/src/async_device/radio.rs [hunk 2]
```diff
@@ -26,6 +24,40 @@ pub trait Timer {
...
+    }
+}
+
+/// An actual reception state
+pub enum RxState {
+    PreambleReceived,
+    Done { length: u8, lq: RxQuality },
+}
+
+#[cfg(feature = "external-lora-phy")]
```
lthiery, line 49: Same as previous

lthiery (changes requested):
Just some small remaining nits, but this is great! Thanks @ilya-epifanov 

I would consider adding a few unit tests to the ToA calculation, but those could be added later as well.

device/src/async_device/mod.rs
```diff
@@ -460,105 +470,136 @@ where
...
-            match select(rx_fut, timeout_fut).await {
-                // RX is complete!
-                Either::Left((Ok((sz, _q)), _)) => Ok(sz),
-                // Timeout or other RX error.
-                _ => Err(Error::RxTimeout),
-            }
-        };
+        let rx_config = self.region.get_rx_config(self.datarate, frame, &Window::_2);
+        #[cfg(feature = "defmt")]
+        trace!("RX2 with config: {}", &rx_config);
```
lthiery, line 589: maybe put these traces behind an additional feature-gate? seems excessive

ilya-epifanov: if DEFMT_LOG is lower than "trace" at compile-time, the whole macro evaluates to nothing

ilya-epifanov:
I've extracted all the changes to the `lora-modulation` into a separate PR #145

lthiery:
> I've extracted all the changes to the `lora-modulation` into a separate PR #145

Good idea! Landed that one second ago

ilya-epifanov:
@lthiery I think we need to publish a new version of `lora-modulation`. I've marked additions to the public API in #145 for convenience

lthiery:
@ilya-epifanov Thanks! We don't have a big precedent for documenting releases yet - if we were breaking APIs, I might bother but let's just get on with it. Can you do me a favor and [bump modulation](https://github.com/ivajloip/rust-lorawan/blob/master/modulation/Cargo.toml#L3) to `0.1.3` in a PR? That way I can land it to main and publish it to crates (I can't self-land PRs).

lthiery:
@ilya-epifanov thanks - 0.1.3 is available now

device/src/async_device/mod.rs
ilya-epifanov: renamed into max_expected_payload_length

ilya-epifanov:
what's funny is that the build succeeded despite `lora-phy` missing the PR https://github.com/embassy-rs/lora-phy/pull/33 
https://github.com/ivajloip/rust-lorawan/actions/runs/6487742044/job/17618701363?pr=142

lthiery:
@ilya-epifanov The async stuff is feature-gated and only build on nightly because async traits are not yet stable. You'll see the nightly build did indeed fail.

lthiery:
With that in mind, do you want to temporarily set Cargo.toml to your branch? We won't land it in master here until lora-phy has an update, but we can at least check that everything is passing.

Also, you don't have to, but it might be cool to write a unit test proving your feature. You should be able to rig it up so that the "preamble is detected" in the mock radio and then you can fire the timer and then deliver the rest of the packet. If the reception is confirmed, it proves your feature out!

ilya-epifanov:
Regarding tests, I think the mock device needs some rework to make testing this possible.
Right now it uses a stateless TX data -> RX data function which will not work if we want to test lower level implementation details.

lthiery:
> Regarding tests, I think the mock device needs some rework to make testing this possible. 

Yes, the mock device will definitely need some rework to test this. I think it's just a question of being able to send a FirePreamble message on the existing mpsc channel. I'll see if I can rig it up quickly

lthiery:
@ilya-epifanov i did not have time to finish, but i think i made most of the necessary changes to make the test possible. PR is against your branch

ilya-epifanov (commented):
I did a self-review. LGTM

device/src/async_device/test/mod.rs
```diff
@@ -34,9 +34,41 @@ async fn test_join_rx1() {
...
+
+    // Trigger beginning of RX1
+    timer.fire().await;
+    // RX preamble
+    radio.fire_preamble();
+    // Preamble received
+    timer.fire().await;
+
+    // Trigger handling of JoinAccept
+    radio.handle_rxtx(handle_join_request);
```
lthiery, line 63: I'm a little confused that this doesn't trigger the panic from `Msg::Preamble, TargetRxState::Done` condition. The fire_preamble should have but us targeting the `TargetRxState::Done` event and handle_rxtx precedes the frame with another preamble... I'll take a look later but wanted to drop the note.

lulf (dismissed):
I can confirm these changes works on my lora-discovery board (although requiring some custom configuration for joining to work in embassy, but it is likely unrelated to this PR since it is required on the main branch too).

I'll leave the final decision to merge when you're happy with the code @lthiery

lthiery:
> I'll leave the final decision to merge when you're happy with the code @lthiery

I think the code is in a good place! I'm just waiting on lora-phy to be settled so we can stay dependent on releases and no forks.

lthiery:
I investigated and found a few things to improve in the test fixtures. [PR'ed](https://github.com/ilya-epifanov/rust-lorawan/pull/2) against @ilya-epifanov 

One thing I'd like to note is that this PR should probably be landed after #144 as it impacts less code and it should be an easier rebase. Furthermore, @ilya-epifanov's PR to `lora_phy` is now on main but until the crates.io namespace is figured out, we cannot depend on a new `lora_phy` release anytime soon. Therefore, I think it's fine for us to depend on the repo & and a hash until the namespace is transferred or until a new namespace is selected. We will simply be blocked from our own releases until that is resolved.

lthiery:
@ilya-epifanov If you are able to do a rebase, we can land this now. If you don't have the time to get a rebase done this week, I can try to take care of it.

plaes:
@ilya-epifanov Seems like it needs another round of rebases :(

plaes:
@ilya-epifanov @lthiery I have a WIP refactored branch here: https://github.com/plaes/rust-lorawan/tree/preamble-detection

ilya-epifanov:
@plaes Thanks a lot. I tried rebasing it but didn't have enough time to finish

lthiery:
@ilya-epifanov thanks so much for the initial inspiration on this. #204 (mostly done by @plaes) has integrated the idea but while offloading the timeout management to the radio. hopefully, that resolves the issue for you now!

PR #144, opened 2023-10-09
Establish Mac module for protocol handling and streamline nb_device + async_device to depend upon it

lthiery:
This refactor simplifies the `nb_device` state machine to only concern itself with radio state. The mac-layer's state is refactored into an orthogonal state machine in the `mac` module. 

The intention is that in a future PR, `async_device` will be able to offload it's own management of the mac layer to this shared `mac` module.

The tests contained in `nb_device/test` demonstrate that functionality is preserved with only a small API change around joining. This change is done to align the nb_device API with that of async_device.


device/src/nb_device/test/mod.rs
```diff
@@ -21,14 +21,15 @@ fn test_join_rx1() {
...
+    let mut device = test_device();
     device.get_radio().set_rxtx_handler(handle_join_request);
-    let response = device.handle_event(Event::NewSessionRequest).unwrap();
+    let response = device.join(get_otaa_credentials()).unwrap();
     assert!(matches!(response, Response::TimeoutRequest(5000)));
     let response = device.handle_event(Event::TimeoutFired).unwrap();
     assert!(matches!(response, Response::TimeoutRequest(5100)));
     // send a timeout for end of rx2
     let response = device.handle_event(Event::TimeoutFired).unwrap();
+    println!("{:?}", response);
```
ivajloip, line 32: is this intentional?

device/src/nb_device/state.rs
```diff
@@ -0,0 +1,397 @@
+/*
+
+This is state machine creates a non-blocking and no-async structure for coordinating radio events
```
ivajloip, line 3: nit: this doesn't read too well - should the `is` be removed?

ivajloip (approved):
I went through all the changes (not in all the details everywhere, but still 🙂) and it looks good to me!

device/src/mac/session.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl Session {
+    pub(crate) fn handle_rx<C: CryptoFactory + Default>(
+        &mut self,
+        region: &mut region::Configuration,
+        configuration: &mut super::Configuration,
+        rx: &mut [u8],
+        dl: &mut Option<Downlink>,
+    ) -> Option<Response> {
```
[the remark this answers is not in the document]
lthiery: I changed this to Response actually. Let me know what you think

lthiery:
Refactor is complete and essentially async_device and nb_device are focused on handling communication between a pure protocol module (Mac) and their radio/timer traits. This means developing mac features benefits both implementations without additional work.

In addition, two issues from async_device are resolved:
* Noise (unparseable lorawan rf) does not upset the rx windows. The rx timeout future is maintained while the buffer is processed (ie: issue #117)
* A confirmed uplink with NoAck is communciated to the user

I could use some testing on hardware if anyone has time. But unit test provide reasonable coverage. I hope to test on hardware myself next weekend.

lthiery:
I've completed testing on hardware and in doing so came across some shortcomings in async's public API. I think it's all resolved now.

Finally, I also tucked away many of the "nb_device" stuff into its own module, so that the structure between async_device and nb_device would be a little more parallel.

There's one thing I'm not certain about at this point and that has to do with the nuance between "Response" and "Error" in the Result<Response, Error>.

I feel like Response is needed to at least differentiate between RxComplete (formerly RxTimeout) and DownlinkReceived. But I've also put common failure modes in there, in particular, when an expected downlink is not received: NoAck and NoJoinAccept.

On one hand, they _are_ failure modes and therefore Errors, so they should be in the Error type. However, I feel like putting them into Response encourages handling on the user side (ie: I expect exhaustive handling of Response). But perhaps that's wrong... Anyone have input?

lulf (approved):

PR #151, opened 2023-10-22
Remove Optional RNG in favor of PRNG to Satisfy RngCore

jbeaurivage:
In my attempts of adapting this library to hardware setups where the user might want to use the TRNG integrated to the LoRa chip (as opposed to more traditional setups where the TRNG is a peripheral of the MCU itself), I wrote #102, and subsequently #109 to allow using the `AsyncRng` trait from lora-phy.

What I ended up with is a complicated typelevel approach that would allow either `Radio` to implement `AsyncRng`, or the user to provide a `RngCore` implementation at compile time. Things were getting a bit out of hand, so I figured there must be another way.

# A new approach based on pseudorandom number generators

After digging for a bit, I realized this crate uses the random number generator for exactly two things:
* Generating `DevNonce`s for join requests
* Selecting random channels when transmitting uplinks.

The good news is that both these operations don't require true cryptographic randomness. In fact, in both cases, we don't even care about predictability! A pseudorandom number generator initialized with a seed generated by a true random number generator is plenty enough:
* `DevNonce`s must only be unique with a low chance of collision. The 1.0.4 LoRaWAN spec even explicitely requires the `DevNonce`s to be a sequence of incrementing integers, which is obviously predictable.
* No one cares if the channel selected  for the next uplink is predictable, as long as the channel selection yields an uniform distribution.

# What's in this PR?

Replace all uses of `RandCore::next_u32` with `fastrand::Rng::u32`. [`fastrand`](crates.io/crate/fastrand/) uses the [wyhash](https://github.com/wangyi-fudan/wyhash) algorithm, which is the default hashing algo for Zig, Go and Nim. We also replace the complex typelevel stuff with a series of simple functions which either accept a reference to a `RngCore` implementation, or a `Radio` implementing `AsyncRng`. We only use these to generate the initial seed, after which `fastrand` does the heavy lifting.

There is also a function provided in case the user would like to provide their own seed, but I'm unsure whether this is too dangerous to include, or if some warnings in the docs would be enough.

Closes #109.


lulf:
I've been thinking about the same a few times , I think it makes sense to do this. Have you measured impact on code size and execution time?( And what hardware?)

jbeaurivage:
Glad to hear it. I can probably check for code size, although I can't really test execution time, as I'm traveling right now and I don't have hardware with me.

I'd be very appreciative if anyone else wants to volunteer to test the PR!

lthiery:
I really like this approach and I assume, if fastrand/wyhash don't cut it, we can find a pseudorandom number generator that fits the bill.

I agree that benchmarking codesize and execution time would be interesting, especially compared to an MCU's built-in RNG.

However, I would suggest the following design which would allow us to not care and we could promptly land this change:
* make it our library internally wants the `async_device/nb_device::Device` structs to contain a type that implement RngCore. ie:
```rust
struct Device<RNG: RngCore> {
   rng: RNG
}
```
* provide an API for using fastrng to fulfill RngCore. Hide this internally and ask for the user to provide a seed to the Device constructor in order to use the library default PRNG. Currently, the design requires passing something that implements `lora_phy::mod_traits::get_random_number` but I'm not sure if there's an advantage as opposed to allowing the user to provide a seed.

With the above we still strip back the complexity of the OptionalRng from radio but we provide flexibility to provide a more optimal RNG (let it be PRNG or using a built-in RNG from the microcontroller). Benchmarking and improving our internal PRNG algorithm can be done after this and it's easy to test and propose alternates as the RNG remains modular (while still providing a simple default that gets users up and going quickly).

jbeaurivage:
That makes a lot of sense, thanks for the review. I'll try to take another pass this evening.

jbeaurivage:
@lthiery, @lulf, I've implemented the proposed changes.

lthiery:
@jbeaurivage at a glance this looks good! I'll give it a more thorough look through in the morning

device/src/async_device/mod.rs
```diff
@@ -84,20 +83,43 @@ enum RxWindowResponse<F: futures::Future<Output = ()> + Sized + Unpin> {
...
+    pub fn new_with_seed_and_session(
         region: region::Configuration,
         radio: R,
         timer: T,
-    ) -> Device<R, C, T, NoneT, N> {
-        Device::new_with_session(region, radio, timer, NoneT, None)
+        seed: u64,
+        session: Option<Session>,
+    ) -> Self {
+        let rng = crate::rng::Prng::new(seed);
```
lthiery, line 121:
```suggestion
        let rng = rng::Prng::new(seed);
```

lthiery (approved):

lthiery:
Tiny correction and then let's land it.

lthiery (approved):

PR #152, opened 2023-10-23
Adapt to lora-phy's BoardType removal

plaes:
Upsteam lora-phy has dropped global BoardType enum in favor of driver-specific configuration.
In rust-lorawan BoardType was being used to specify board-specific RX1/RX2 window timing tweaks.

### DIFF
--- a/lorawan-device/src/async_device/lora_radio.rs
+++ b/lorawan-device/src/async_device/lora_radio.rs
@@ -1,7 +1,7 @@
 use super::radio::{PhyRxTx, RfConfig, RxQuality, TxConfig};
 use super::Timings;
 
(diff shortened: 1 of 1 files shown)


lulf (commented):
Changes look "ok" if the rx windows logic is being improved down the line, iirc this moves it back to how it was so I think this needs to be tested on the different boards in the list too.

plaes:
I have split the chip-specific power handling into its own PR so hopefully things progress faster this way.

lthiery:
Perhaps defaults are sensible but allow customization by user?

Dirbaio:
merged as part of https://github.com/lora-rs/lora-rs/pull/172

PR #158, opened 2023-10-28
Class C

lthiery:


device/src/async_device/mod.rs
```diff
@@ -237,6 +237,10 @@ where
     /// The returned future completes when the data have been sent successfully and downlink data,
     /// if any, is available by calling take_downlink. Response::DownlinkReceived indicates a
     /// downlink is available.
+    ///
+    /// In Class C mode, it is possible to get one or more downlinks and `Reponse::DownlinkReceived`
+    /// maybe not even be indicated. It is recommended to call `take_downlink` after `send` until
+    /// it ruturns `None`.
```
plaes, line 269: s/ruturns/returns

device/src/async_device/mod.rs
```diff
@@ -34,7 +35,7 @@ use core::cmp::min;
 /// - C: A CryptoFactory implementation
 /// - RNG: A random number generator implementation. An external RNG may be provided, or you may use a builtin PRNG by
 /// providing a random seed.
-pub struct Device<R, C, T, G, const N: usize = 256>
+pub struct Device<R, C, T, G, const N: usize = 256, const D: usize = 4>
```
lulf, line 38: Probably makes sense to default to 1 to avoid accidentally increasing memory for existing users

device/src/test_util.rs
```diff
@@ -64,14 +65,20 @@ pub fn get_abp_credentials() -> JoinMode {
...
 pub type RxTxHandler = fn(Option<Uplink>, RfConfig, &mut [u8]) -> usize;
 
+lazy_static::lazy_static! {
+    static ref SESSION: Mutex<HashMap<usize, Session>> = Mutex::new(HashMap::new());
+
+}
+
 /// Handle join request and pack a JoinAccept into RxBuffer
-pub fn handle_join_request(
+pub fn handle_join_request<const I: usize>(
```
lulf, line 74: I think it would make sense to have this return Result() than add so many panic!'s

lulf (commented):
Thanks 🚀  Left some inline comments.

device/src/test_util.rs
plaes: @lulf It's a helper for tests

lulf: Ah, good to know, thanks!

device/src/async_device/mod.rs
lthiery: Good point! Thanks

lthiery:
I'm considering removing the loop after Join and instead informing the user that an uplink should be sent after join before Class C Downlinks may be received.

In addition, I think the `enable_class_c shall()` should not actually touch the radio and only set the flag.

lorawan-device/src/async_device/mod.rs
```diff
@@ -254,7 +281,64 @@ where
...
+            self.radio.low_power().await.map_err(Error::Radio)?;
+        }
+        Ok(())
+    }
+
+    async fn between_windows(
+        &mut self,
+        duration: u32,
+    ) -> Result<Option<mac::Response>, Error<R::PhyError>> {
+        if self.class_c {
```
plaes, line 301: Could you switch the check to deal with `if !self.class_c` as this would allow to do an early return and reduce the indent level for the "big part" of this function...
Once code for class B arrives, this "flag" could be turned to optional enum (None, Some(class_b), Some(class_c)...

lorawan-device/src/mac/mod.rs
```diff
@@ -223,6 +228,27 @@ impl Mac {
         }
     }
 
+    /// Handles a received RF frame during RXC window. Returns None is unparseable, fails decryption,
```
plaes, line 231: Returns `None` if frame is unparseable or fails either decryption or MIC verification. (And I don't really understand the success case...).

lthiery: In English, saying `A, B, or C` is equivalent to saying `A||B||C`. I think it's fine as is on that end?

However, I've tried to clarify the success case better.

lorawan-device/src/async_device/mod.rs
lthiery: Thanks, I like the idea! Not that it matters, but I'm skeptical that Class B will ever play into this section though.

lorawan-device/src/mac/mod.rs
plaes: The **is** after None should be 'if'?

lthiery: Oh right on about that! I thought you were emphasizing the ORs - thanks

PR #159, opened 2023-11-05
extend PhyTxRx to consider antenna gain and max power

### LINKED ISSUE #150: DBM values in Australia and America block usage of low-power ChipType::SX1261
AU_DBM and US_DBM are set to 21, which is beyond the 15 dBm maximum when using ChipType::SX1261, including STM32WLxx configured for the Low-Power Amplifier. Calling lora_phy's set_pa_config when ChipType::SX1261 and the Australia or US regions returns a result of Err(RadioError::InvalidOutputPower).

Need a way to pass in a different value, or change the default value when using ChipType::SX1261.

lthiery:
Resolves #150

### DIFF
--- a/lorawan-device/src/async_device/lora_radio.rs
+++ b/lorawan-device/src/async_device/lora_radio.rs
@@ -6,8 +6,11 @@ use lora_phy::mod_params::{BoardType, ChipType, RadioError};
 use lora_phy::mod_traits::RadioKind;
 use lora_phy::{DelayUs, LoRa};
 
-/// LoRa radio using the physical layer API in the external lora-phy crate
-pub struct LoRaRadio<RK, DLY>
+/// LoRa radio using the physical layer API in the external lora-phy crate.
+///
+/// The const generic P is the max power the radio may be instructed to transmit at. The const
+/// generic G is the antenna gain and board loss in dBi.
+pub struct LoRaRadio<RK, DLY, const P: u8, const G: i8 = 0>
 where
     RK: RadioKind,
     DLY: DelayUs,
@@ -15,7 +18,7 @@ where
     pub(crate) lora: LoRa<RK, DLY>,
     rx_pkt_params: Option<lora_phy::mod_params::PacketParams>,
 }
-impl<RK, DLY> LoRaRadio<RK, DLY>
+impl<RK, DLY, const P: u8, const G: i8> LoRaRadio<RK, DLY, P, G>
 where
     RK: RadioKind,
     DLY: DelayUs,
@@ -26,7 +29,7 @@ where
 }
 
 /// Provide the timing values for boards supported by the external lora-phy crate
-impl<RK, DLY> Timings for LoRaRadio<RK, DLY>
+impl<RK, DLY, const P: u8, const G: i8> Timings for LoRaRadio<RK, DLY, P, G>
 where
     RK: RadioKind,
     DLY: DelayUs,
@@ -63,12 +66,16 @@ impl From<RadioError> for Error {
 
 /// Provide the LoRa physical layer rx/tx interface for boards supported by the external lora-phy
 /// crate
-impl<RK, DLY> PhyRxTx for LoRaRadio<RK, DLY>
+impl<RK, DLY, const P: u8, const G: i8> PhyRxTx for LoRaRadio<RK, DLY, P, G>
 where
     RK: RadioKind,
     DLY: DelayUs,
 {
     type PhyError = Error;
+
+    const ANTENNA_GAIN: i8 = G;
+
+    const MAX_RADIO_POWER: u8 = P;
 
     async fn tx(&mut self, config: TxConfig, buffer: &[u8]) -> Result<u32, Self::PhyError> {
         let mdltn_params = self.lora.create_modulation_params(
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -147,7 +147,7 @@ where
         rng: G,
         session: Option<Session>,
     ) -> Self {
-        let mut mac = Mac::new(region);
+        let mut mac = Mac::new(region, R::MAX_RADIO_POWER, R::ANTENNA_GAIN);
         if let Some(session) = session {
             mac.set_session(session);
         }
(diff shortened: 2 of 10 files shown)


device/src/lib.rs
```diff
@@ -54,6 +54,27 @@ pub trait Timings {
     fn get_rx_window_duration_ms(&self) -> u32;
 }
 
+pub trait BoardEirp {
+    /// Returns the antenna gain in dBi. You should take into account any board losses here.
+    fn get_antenna_gain(&self) -> u8;
```
plaes, line 59: Instead of getter methods, can't we use associated constants?
https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants

lulf (commented):
Why couldn't this be just added to PhyRxTx with default impls to avoid the extra generic arg?

device/src/lib.rs
lthiery: > Also, antenna gain could theoretically be negative..
That's a good point - thanks!

> Instead of getter methods, can't we use associated constants?
Yes, I think that's a good point. I thought maybe getters would be nice because it means it could vary during runtime, but I realized I can't think of a practical use-case for it yet, so let's just start with constants.

Writing the panic by hand seems a bit unnecessary though. I think the normal compiler error does a fine job?
```
35 |     const MAX_RADIO_POWER: u8;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^-
   |                              |
   |                              help: provide a definition for the constant: `= <expr>;
```

lthiery:
> Why couldn't this be just added to PhyRxTx with default impls to avoid the extra generic arg?

Thanks for the feedback! I think you're right and I tried it out combined with @plaes 's const generic suggestion.

device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -58,6 +58,7 @@ impl<const D: usize, F: FixedChannelRegion<D>> FixedChannelPlan<D, F> {
         self.channel_mask.set_bank(7, mask);
     }
 
+    #[allow(unused)]
```
plaes, line 61: `#[cfg(test)]` ?

device/src/async_device/lora_radio.rs
```diff
@@ -6,16 +6,19 @@ use lora_phy::mod_params::{BoardType, ChipType, RadioError};
...
+pub struct LoRaRadio<RK, DLY, const P: u8, const G: i8 = 0>
 where
     RK: RadioKind,
     DLY: DelayUs,
 {
     pub(crate) lora: LoRa<RK, DLY>,
     rx_pkt_params: Option<lora_phy::mod_params::PacketParams>,
 }
-impl<RK, DLY> LoRaRadio<RK, DLY>
+impl<RK, DLY, const G: i8, const P: u8> LoRaRadio<RK, DLY, P, G>
```
plaes, line 21: Here, the `G` and `P` are in different order.

device/src/async_device/radio.rs
```diff
@@ -29,6 +29,13 @@ pub trait PhyRxTx: Sized {
     #[cfg(not(feature = "defmt"))]
     type PhyError;
 
+    /// The antenna gain and board loss in dBi.
+    const ANTENNA_GAIN: i8 = 0;
+
+    /// The max power the radio can be instructed to transmit at. When preparing an instruction
```
plaes, line 35: Something like this? (Also if accepted then same comments should be used for non-async radio)
```rust
/// Maximum power the radio is able to deliver (in dBm?).
/// When preparing instructions for radio, the value of maximum power will be used as an upper bound.
```

plaes (commented):
What I do not like with current design is that it's somewhat error-prone as it has to be "manually" adjusted every time `mac.{send,join_otaa}` creates TxConfig object.

Ideally it should be already adjusted whenever this object is craeted which happens in only two places:
* join_otaa
* send
Or alternatively it should happen when this instance is used (one possibility could be adding extra optional argument for  `radio.tx(...)` in lora-phy)?

device/src/async_device/lora_radio.rs
lthiery: Good catch! Thank you

device/src/async_device/radio.rs
```diff
@@ -29,6 +29,13 @@ pub trait PhyRxTx: Sized {
     #[cfg(not(feature = "defmt"))]
     type PhyError;
 
+    /// The antenna gain and board loss in dBi.
```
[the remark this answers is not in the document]
lthiery: Ty

device/src/async_device/radio.rs
lthiery: I agree - thanks

lthiery:
> What I do not like with current design is that it's somewhat error-prone as it has to be "manually" adjusted every time `mac.{send,join_otaa}` creates TxConfig object.

Thanks for the idea @plaes. Your point is why my first iteration passed a const generic to the Mac struct initially, but @lulf 's feedback made me think that the mental load of yet another trait seemed unnecessary and I liked the UX of the trait associated const.

I was _going_ to do `Mac<R::MAX_RADIO_POWER,R::ANTENNA_GAIN>` but that depends on `generic_const_exprs` which can only be enabled in nightly. Even with that feature enabled, I couldn't quite get it to work and the compiler was telling me that the expressions were not properly constrained...

Therefore, the current solution is to use the trait constants to instantiate the Mac with the appropriate limits. This at least passes the responsibility the adjustment to `join_otaa` and `send` making it harder to make a mistake in the future.

device/src/mac/mod.rs
```diff
@@ -125,6 +133,9 @@ impl Mac {
         let mut otaa = otaa::Otaa::new(credentials);
         let dev_nonce = otaa.prepare_buffer::<C, RNG, N>(rng, buf);
         self.state = State::Otaa(otaa);
+        let mut tx_config =
+            self.region.create_tx_config(rng, self.configuration.data_rate, &Frame::Join);
+        tx_config.adjust_power(self.board_eirp.max_power, self.board_eirp.antenna_gain);
         (self.region.create_tx_config(rng, self.configuration.data_rate, &Frame::Join), dev_nonce)
```
plaes, line 139: This looks odd here. looks like `tx_config` is updated with board_eirp data but then not used.

lthiery: Good catch again! Thanks

plaes (approved):

plaes (approved):

lulf (dismissed):

PR #166, opened 2023-11-14
external-lora-phy: Drop removed ChipType and clamp tx power

plaes:
As lora-phy has been refactored to drop global chiptype enum, this needs to be handled here as well.

There's still some work needed from lora-phy side, so each driver clamps the tx power value to max supported, instead of failing with `InvalidPower` error.

### DIFF
--- a/lorawan-device/src/async_device/lora_radio.rs
+++ b/lorawan-device/src/async_device/lora_radio.rs
@@ -1,8 +1,7 @@
 use super::radio::{PhyRxTx, RfConfig, RxQuality, TxConfig};
-use super::region::constants::DEFAULT_DBM;
 use super::Timings;
 
-use lora_phy::mod_params::{BoardType, ChipType, RadioError};
+use lora_phy::mod_params::{BoardType, RadioError};
 use lora_phy::mod_traits::RadioKind;
 use lora_phy::{DelayUs, LoRa};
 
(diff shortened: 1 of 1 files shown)


lulf (dismissed):

lthiery:
Surprising there's no breakage in the API. Will people running lora-phy = 2.1.1 be impacted by this change?

lthiery:
It looks like the only impact would be that additional chips (`ChipType::CustomChip, ChipType::Sx1261, ChipType::SX1276`) would be attributed this threshold.

Seems fine!

lthiery (dismissed):

lorawan-device/src/async_device/lora_radio.rs
```diff
@@ -86,16 +86,17 @@ where
...
-                }
-            }
-            _ => config.pw,
-        };
+        let pw = config.pw.min(DEFAULT_DBM);
+
+        #[cfg(feature = "defmt")]
+        if config.pw > DEFAULT_DBM {
+            defmt::warn!(
+                "Limiting requested tx power {} with region default: {}",
```
lthiery, line 94: Is it `region default` or `radio default max`?

plaes: Is there a way to get region default from this location?

plaes:
Ok, I went over the list of maximum EIRP per region, based on [RP002-1.0.3 Regional Parameters](https://lora-alliance.org/wp-content/uploads/2021/05/RP002-1.0.3-FINAL-1.pdf)
1. EU863-870: +16dBm
2. US902-928: +30dBm (?)
3. CN779-787: +12dBm
4. EU433: +12dBm
5. AU915-928: +30dBm
6. CN470-510: +19dBm
7. AS923: +16dBm
8. KR920-923: (depends on frequency: +12dBm if f < 922MHz), otherwise +14dBm 
9. IN865-867: +30dBm
10. RU864-870: +16dBm

lorawan-device/src/async_device/lora_radio.rs
lthiery: I can't think of an easy way, or a way at all to be honest.

lthiery (approved):

lthiery:
> **[Edit]:** Apparently, this isn't the correct location to handle these, instead should be handled whereever `Region::get_dbm()` is called


> **[Edit]:** Apparently, this isn't the correct location to handle these, instead should be handled whereever `Region::get_dbm()` is called...
> 
> > Ok, I went over the list of maximum EIRP per region, based on [RP002-1.0.3 Regional Parameters](https://lora-alliance.org/wp-content/uploads/2021/05/RP002-1.0.3-FINAL-1.pdf)
> > 
> > 1. EU863-870: +16dBm
> > 2. US902-928: +30dBm (?)
> > 3. CN779-787: +12dBm
> > 4. EU433: +12dBm
> > 5. AU915-928: +30dBm
> > 6. CN470-510: +19dBm
> > 7. AS923: +16dBm
> > 8. KR920-923: (depends on frequency: +12dBm if f < 922MHz), otherwise +14dBm
> > 9. IN865-867: +30dBm
> > 10. RU864-870: +16dBm

Agreed. I don't think many of these have been added to the specific region plans in this repo yet. Only US_DBM has a regionalized one and everyone is defaults to DEFAULT_DBM = 14.

PR #173, opened 2023-11-29
Add the examples from the embassy repo.

Dirbaio:
I've ported the examples from the Embassy repo and updated them to the new API.

:warning:  **I have only checked that they build**! :warning:  I don't have the hardware to run them. In particular, I'd appreciate double-checking that the new `Config` values are right for the boards the examples are supposed to target.


lulf (commented):
Changes look good, i will test the examples i have the devkit for.

plaes:
> Testing on my stm32wl it is able to transmit packets, although it fails to join the network. Past experience is that if something HAL-specific is badly setup, radio won't work at all, and the fact that I keep having RX issues for a long time, suggests that something's up with timing at my end.

@lulf Yeah, there is known issue receiving packet during RX1 window if transmission takes more than 1 second (which default at least in EU868 region).

IMHO, we should just merge this, but add additional information about the caveats (not all boards are tested) and may be broken.

PR #175, opened 2023-12-01
use enums instead of &str for encoding errors

lucasgranberg:
By using enums everywhere error handling should be easier in the long run. This also allows better use of defmt.

### DIFF
--- a/lorawan-device/src/test_util.rs
+++ b/lorawan-device/src/test_util.rs
@@ -1,11 +1,11 @@
 use super::*;
 use lorawan::maccommands::{ChannelMask, SerializableMacCommand};
-use lorawan::parser::DataHeader;
+use lorawan::parser::{self, DataHeader};
 use lorawan::{
     default_crypto::DefaultFactory,
     maccommandcreator::LinkADRReqCreator,
     maccommands::{LinkADRReqPayload, MacCommand},
-    parser::{parse, DataPayload, FCtrl, JoinAcceptPayload, PhyPayload},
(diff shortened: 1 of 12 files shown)


lorawan-device/src/test_util.rs
```diff
@@ -1,11 +1,11 @@
 use super::*;
 use lorawan::maccommands::{ChannelMask, SerializableMacCommand};
-use lorawan::parser::DataHeader;
+use lorawan::parser::{self, DataHeader};
 use lorawan::{
     default_crypto::DefaultFactory,
     maccommandcreator::LinkADRReqCreator,
     maccommands::{LinkADRReqPayload, MacCommand},
-    parser::{parse, DataPayload, FCtrl, JoinAcceptPayload, PhyPayload},
+    parser::{parse, DataPayload, JoinAcceptPayload, PhyPayload},
```
plaes, line 8: `FCtrl` is needed..

Try running tests with `-F async,external-lora-phy`

lucasgranberg: I put it behind the async feature now

lora-phy/src/sx1261_2/radio_kind_params.rs
```diff
@@ -219,6 +220,7 @@ pub enum TcxoCtrlVoltage {
     Ctrl3V3 = 0x07,
 }
 
+#[allow(missing_docs)]
```
plaes, line 223: Is this really needed?

lthiery (approved):

PR #176, opened 2023-12-02
set use_dio2_as_rfswitch for examples

lthiery:
From [RAK documentation](https://docs.rakwireless.com/Product-Categories/WisBlock/RAK4631/Datasheet/#hardware):

"
Important for successful SX1262 initialization:

* Setup DIO2 to control the antenna switch
* Setup DIO3 to control the TCXO power supply
* Setup the SX1262 to use it's DCDC regulator and not the LDO
* RAK4630 schematics show GPIO P1.07 connected to the antenna switch, but it should not be initialized, as DIO2 will do the control of the antenna switch

"


lthiery:
Waveshare https://www.waveshare.com/wiki/SX1262_XXXM_LoRaWAN/GNSS_HAT

PR #177, opened 2023-12-02
typed keys for lorawan-encoding

lthiery:


plaes:
> > One of the annoyances regarding OTAA that I encountered is that `DevEUI` and `AppEUI` are LSB, but `AppKey` is MSB. Maybe it's worth mentioning somewhere?
> 
> Is this in the specs or network specific? Would be great if it could be made more clear as it is easy to get wrong

I'm only familiar with Chirpstack which describes the device provisioning here - https://www.chirpstack.io/docs/chirpstack/use/devices.html

I haven't set up anything with Things Network, and I haven't been able to find anything which mentions LSB/MSB in their docs either - closest what I have found is this page: https://www.thethingsindustries.com/docs/devices/adding-devices/manual/otaa/

lthiery:
FWIW, I think this is due to how the crypto libraries are used and so we lay them out internally to match their usage. This matches the example stack as far as I know. For some example, the hex representation tends to be... the way it is.

Anyway, I think some docs are good to have here!

PR #178, opened 2023-12-02
split maccommands by uplink and downlink

lucasgranberg:
Splitting the maccommands makes sense to me as very few applications needs parsinig and creating both kinds at the same time.

Reacting to maccommands is also easier as you don't need to use a catch all in the match.


lucasgranberg:
tl;dr
- put uplink and downlink in different enums
- move some functionality to a new macro
- change the rest of the monorepo to work with this change

lulf (commented):
Seems like a reasonable change to me! Since it's a draft, are you planning more changes?

lucasgranberg:
> Seems like a reasonable change to me! Since it's a draft, are you planning more changes?

I want to see if the macro and resulting structs actually makes sense. The focus has been on the split more than the result so there are probably some things to tidy up.  I will work on it tomorrow to tidy it up and then make it in to an actual PR

lorawan-encoding/src/maccommands.rs
```diff
@@ -196,23 +85,27 @@ macro_rules! mac_cmds {
...
         $(
             $(#[$outer])*
-            pub struct $type<'a>(&'a [u8]);
+            pub struct $type<'a>(pub(crate) &'a [u8]);
 
             impl<'a> $type<'a> {
                 /// Creates a new instance of the mac command if there is enought data.
                 pub fn new(data: &'a [u8]) -> Result<$type<'a>, Error> {
-                    if data.len() < $size {
+                    if data.len() != $size {
```
ivajloip, line 93: The error type might be misleading if the size is bigger.

lorawan-encoding/src/maccommands.rs
```diff
@@ -222,21 +115,154 @@ macro_rules! mac_cmds {
...
+                /// Reference to the payload.
+                pub fn bytes (&self) -> &[u8]{
+                    self.0
+                }
             }
         )*
+    }
+}
+
+macro_rules! mac_cmds_enum {
```
ivajloip, line 132: I am not sure if I fully understood this, but it seems beautiful! 😃

lorawan-encoding/src/parser.rs
```diff
@@ -1059,12 +1060,13 @@ impl<'a> FHDR<'a> {
...
-        let f_opts_len = FCtrl(self.0[4], self.1).f_opts_len();
-        let start = 7;
-        let end = 7 + f_opts_len as usize;
-        parse_mac_commands(&self.0[start..end], self.1)
+    /// Gives the size of FOpts.
+    pub fn fopts_len(&self) -> u8 {
+        FCtrl(self.0[4], self.1).f_opts_len()
+    }
+
+    pub fn data(&self) -> &[u8] {
```
ivajloip, line 1068: I am thinking - if I read just the function name, I will not be sure what this returns, so maybe a short comment with that will be helpful.

lorawan-encoding/src/parser.rs
```diff
@@ -1130,8 +1132,12 @@ impl<'a> FRMMacCommands<'a> {
...
-    pub fn mac_commands(&self) -> MacCommandIterator {
-        parse_mac_commands(self.1, self.0)
+    pub fn data(&self) -> &[u8] {
+        self.1
     }
+
+    // /// Gives the list of mac commands represented in the FRMPayload.
+    // pub fn mac_commands(&self) -> MacCommandIterator {
+    //     parse_mac_commands(self.1, self.0)
+    // }
```
ivajloip, line 1142: I guess this is going to be deleted when you finish this.

ivajloip (commented):
It looks good so far! 🙂

lorawan-encoding/src/maccommands.rs
lucasgranberg: Yes, I don't know what I was thinking there :thinking: 
changing it to < would allow somebody to create the mac command from a larger buffer, leaving it as is should at least change the error message.

For now I will change it back to how it was with <

lorawan-encoding/src/maccommands.rs
lucasgranberg: heh thanks, macros can make it hard to read but basically I pass an enum to a macro and move some of the existing fns to the macro to avoid writing them for every maccommand. Here some of the fns could probably be inlined. Some places the fns require self even when it is not needed bc it needs to have the same signture. This is especially true for zero length commands

lorawan-encoding/src/parser.rs
lucasgranberg: Right you are, I did some cleaning here. I also removed the fopts_len fn as it duplicates what already happens in FCtrl

lucasgranberg:
I had some more stuff I wanted to add but that stuff can be added later in a separate PR as it goes beyond the scope of splitting the mac commands. Some of the stuff is needed to make 1.0.4 working (ex. ADR). For now I think this is starting to become ready to review/merge

lthiery (approved):

lthiery:
This looks great! And it looks complete in itself so a good PR. More can come in future PRs!

PR #180, opened 2023-12-04
Update Embassy, remove embedded-storage patch.

Dirbaio:
Embedded-storage v0.3.1 and embedded-storage-async v0.4.1 are now released, so the patch can be removed.
These were the last deps using `#[feature(...)]`, so all deps now work on beta (and the upcoming 1.75 stable!).


Dirbaio:
fixed :upside_down_face:

lthiery (approved):

PR #181, opened 2023-12-04
consistently store channels as u8. flow fix to reduce bugs

lthiery:
I noticed on the first data uplink for US915, the appropriate downlink channel is not selected. This fixes the logic (which was very similar to the bug fixed in #160) and tries to structure the code to make it less error prone.

### DIFF
--- a/lorawan-device/src/region/fixed_channel_plans/join_channels.rs
+++ b/lorawan-device/src/region/fixed_channel_plans/join_channels.rs
@@ -36,7 +36,7 @@ impl JoinChannels {
                 self.previous_channel % 8
             };
             // pick another channel on that subband
-            Some((rng.next_u32() as u8 % 8) + (sb * 8))
+            Some((rng.next_u32() & 0b111) as u8 + (sb * 8))
         } else {
             None
         }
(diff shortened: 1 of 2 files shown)


lorawan-device/src/region/fixed_channel_plans/join_channels.rs
```diff
@@ -56,19 +58,19 @@ impl JoinChannels {
...
 
-    pub(crate) fn get_next_channel(&mut self, rng: &mut impl RngCore) -> usize {
+    pub(crate) fn get_next_channel(&mut self, rng: &mut impl RngCore) -> u8 {
         match (self.preferred_subband, self.num_retries.cmp(&self.max_retries)) {
             (Some(sb), Ordering::Less) => {
                 self.num_retries += 1;
                 // pick a  random number 0-7 on the preferred subband
                 // NB: we don't use 500 kHz channels
-                let channel = (rng.next_u32() as usize % 8) + ((sb as usize - 1) * 8);
+                let channel = (rng.next_u32() as u8 % 8) + ((sb as usize - 1) as u8 * 8);
```
plaes, line 67: `rng.next_u32() as u8` feels quite shady. Can you try with a sample program that prints out bunch of  downcasted values?

lthiery: nice catch! i totally forgot about that
https://doc.rust-lang.org/rust-by-example/types/cast.html

lorawan-device/src/region/fixed_channel_plans/join_channels.rs
```diff
@@ -64,7 +64,7 @@ impl JoinChannels {
                 self.num_retries += 1;
                 // pick a  random number 0-7 on the preferred subband
                 // NB: we don't use 500 kHz channels
-                let channel = (rng.next_u32() as u8 % 8) + ((sb as usize - 1) as u8 * 8);
+                let channel = ((rng.next_u32() & 0xFF) as u8 % 8) + ((sb as usize - 1) as u8 * 8);
```
plaes, line 67: Not sure whether there's point need to clamp it as `32008 as u8` will be 8 anyway.
So this could be just simply `(rng.next_u32() % 8) as u8 + ...`

PR #185, opened 2023-12-24
bump embassy version of examples

lthiery:


lulf (approved):

lthiery:
i was trying to fix CI after landing #184 but I couldn't find a version of embassy-time that matched. I think it could be fixed by patching the embassy release that embassy-time 0.2.0 relies on, but I figured we may as well wait for rust 1.76 which I imagine when the embassy 1.0.0 is official?

PR #187, opened 2023-12-27
Bump ehal

CBJamo:
Draft pending the various upstream movements. Will need to remove .cargo/config.toml


lthiery:
Replaced by #201

PR #189, opened 2023-12-29
refactor `external-lora-phy` in `lorawan-device` as `lorawan-radio` in `lora-phy`

lthiery:
It's my opinion that it's better practice to place trait implementations of a struct within the same crate/module as the struct definition itself. In either case, the feature remains feature-gated, however, it feels more natural to have a feature "add a trait implementation to an existing struct" rather than "import a struct an implement the feature with it".


lucasgranberg (dismissed):
There are pros and cons with both approaches but I think the way you suggest is better. This also allows for lorawan specific stuff to be implemented on the radio if needed.

lucasgranberg:
I did not notice that lorawan-device sneaked in as a dependency.

```toml
lorawan-device = { path = "../lorawan-device", version = ">=0.11.0", features = [
    "defmt",
], optional = true }
...
[features]
lorawan-radio = ["dep:lorawan-device"]

```

plaes:
> I did not notice that lorawan-device sneaked in as a dependency.
> 
> ```toml
> lorawan-device = { path = "../lorawan-device", version = ">=0.11.0", features = [
>     "defmt",
> ], optional = true }
> ...
> [features]
> lorawan-radio = ["dep:lorawan-device"]
> ```
I think it should be `lorawan-device/async` ?

lucasgranberg:
> I think it should be `lorawan-device/async` ?

Well for only fetching lorawan-device only when needed my suggestion should work
Only place lorawan-device is used is in lorawan_radio.rs and that mod is behind the lorawan-radio feature.

that being said you are right. Missing async feature for lorawan-device seems to be an other issue, I don't know how this works without it

lucasgranberg:
https://github.com/lora-rs/lora-rs/pull/195

PR #190, opened 2023-12-29
update aes and cmac libraries

lthiery:


plaes (dismissed):

lorawan-encoding/Cargo.toml
```diff
@@ -10,9 +10,9 @@ keywords = ["lorawan", "iot", "lpwan", "parser", "lightweight"]
 readme = "README.md"
 
 [dependencies]
-aes = { version = "0.6.0", optional = true }
-cmac = { version = "0.5.1", optional = true }
-generic-array = "0.14"
+aes = { version = "0.8", optional = true }
+cmac = { version = "0.7", optional = true }
+generic-array = "0"
```
lucasgranberg, line 15: Do we need generic-array if we can use the one bundled with aes?

lthiery: We need it for defining the generic Crytofactory trait. aes/cmac comes in just for DefaultFactory.

lucasgranberg (dismissed):
lgtm

PR #193, opened 2024-01-02
allow fine-tuning of timing for lorawan-device

lthiery:


plaes:
Could you try building with the `#![allow(missing_docs)]` removed on the first line?

lthiery:
> Could you try building with the #![allow(missing_docs)] removed on the first line?

I added that when I moved the module from lorawan-device to lora-phy. There are some undocumented structs...

lora-phy/src/lorawan_radio.rs
```diff
@@ -29,22 +34,36 @@ where
...
+}
+
 /// Provide the timing values for boards supported by the external lora-phy crate
 impl<RK, DLY, const P: u8, const G: i8> Timings for LorawanRadio<RK, DLY, P, G>
 where
     RK: RadioKind,
     DLY: DelayNs,
 {
-    // TODO: Use a struct with Default or Option fallback?
     fn get_rx_window_offset_ms(&self) -> i32 {
```
lucasgranberg, line 62: Could you add a comment on what this offset and duration actually does?

lthiery: The comment exists on the trait definition already: https://github.com/lora-rs/lora-rs/blob/main/lorawan-device/src/lib.rs#L61-L70

lucasgranberg: Whoops, didn't see that

lucasgranberg (approved):

PR #194, opened 2024-01-02
make lorawan-encoding support 1.0.4 

lucasgranberg:
This cleans up maccommandcreator and adds ADR fields.

### DIFF
--- a/lorawan-encoding/src/maccommandcreator.rs
+++ b/lorawan-encoding/src/maccommandcreator.rs
@@ -38,6 +38,16 @@ macro_rules! impl_mac_cmd_creator_boilerplate {
             pub fn build(&self) -> &[u8] {
                 &[$cid]
             }
+
+            /// Get the CID.
+            pub const fn cid(&self) -> u8 {
+                $cid
+            }
+
+            /// Get the length.
+            pub const fn len(&self) -> usize {
+                1
+            }
         }
 
         impl_mac_cmd_payload!($type);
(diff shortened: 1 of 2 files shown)


lorawan-encoding/src/maccommandcreator.rs
```diff
@@ -612,11 +660,13 @@ impl RXTimingSetupReqCreator {
 /// let creator = lorawan::maccommandcreator::RXTimingSetupAnsCreator::new();
 /// let res = creator.build();
 /// ```
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct RXTimingSetupAnsCreator {}
 
 impl_mac_cmd_creator_boilerplate!(RXTimingSetupAnsCreator, 0x08);
 
-#[derive(Debug, PartialEq, Eq)]
```
lthiery, line 619: Is there reasoning in removing PartialEq, Eq?

lorawan-encoding/src/parser.rs
```diff
@@ -1083,6 +1083,16 @@ impl FCtrl {
         FCtrl(bytes, uplink)
     }
 
+    /// Set ADR ACK requested.
+    pub fn set_adr_ack_req(&mut self) {
+        self.0 |= 0b1 << 6;
```
plaes, line 1088: Just use `1 << 6`,

lorawan-encoding/src/parser.rs
```diff
@@ -1083,6 +1083,16 @@ impl FCtrl {
...
+    }
+
+    /// Set ADR enabled.
+    pub fn set_adr(&mut self) {
+        self.0 |= 0b1 << 7;
```
plaes, line 1093: Ditto.

lorawan-encoding/src/maccommandcreator.rs
lucasgranberg: my reasoning was that they were implemented only for some commands and not all. I also don't see a reason for comparing two maccommandcreator objects. I can add them back if needed, Ill try to check in git blame why they are there in the first place.

lucasgranberg: Turns out I am to blame. I added them half-assedly when I first added the 1.0.4 commands.

PR #195, opened 2024-01-03
lora-phy: put lorawan-device behind lorawan-radio feature, add async

lucasgranberg:
This removes lorawan-device dependency for people not needing it
The async feature is needed to make this compile


plaes (approved):
Thanks! :100:

PR #197, opened 2024-01-03
update embedded-hal* to 1.0.0-rc.3

lucasgranberg:


lucasgranberg:
This approach might be too naive. embassy would need updating too plus there is a PR that touches this. Will continue tomorrow
https://github.com/lora-rs/lora-rs/pull/187

PR #198, opened 2024-01-04
Support for preamble detection during RX1-RX2 windows

### LINKED ISSUE #115: Avoid RX1 cancellation during RF reception
```
3.3.1 Receiver activity during receive windows
If a preamble is detected during one of the receive windows, the radio receiver SHOULD stay
active until the downlink frame is demodulated. If a frame was detected and subsequently
demodulated during the first receive window, and the frame was intended for this end-device
after address and MIC (message integrity code) checks, the end-device SHALL NOT open
the second receive window.
```

The code [here](https://github.com/ivajloip/rust-lorawan/blob/7f7ceef024991d0f068b59cff715897b021e8521/device/src/state_machines/session.rs#L502C18-L502C18) should handle `radio::Response::Rxing` appropriately.

plaes:
This is a draft PR for receiving packets that take longer to receive than RX1 and RX2 windows.
Obsoletes PR #142 and once properly implemented, fixes #115.

- [x] Figure out issue with RX2 not working - Apparently RX2 packet is only pushed out if gateway runs into scheduling problems during RX1.
- [x] Test with sx127x radios
- [x] Configure timeout for both preamble and packet payload reception
- [x] Clean up window start and duration calculations
- [ ] Fix Class C tests (test runner needs to handle preamble properly)

Currently tested within EU868:
* sx1262 -RAK4631 with nrf52840
* sx1272 - custom board with nrf52832

Things left for upcoming PRs:
* ValidHeader IRQ handling to detect whether packet is actually intended for our device by checking address + MIC
* Preamble noise handling (ie currently when we detect preamble during RX1, we won't start RX2 window)
* Refactor RX path for Class A to not use continuous RX

### DIFF
--- a/lora-phy/src/lorawan_radio.rs
+++ b/lora-phy/src/lorawan_radio.rs
@@ -1,10 +1,10 @@
 #![allow(missing_docs)]
 
-use lorawan_device::async_device::radio::{PhyRxTx, RfConfig, RxQuality, TxConfig};
+use lorawan_device::async_device::radio::{PhyRxTx, RfConfig, RxQuality, RxState, TargetRxState, TxConfig};
 use lorawan_device::Timings;
 
 use super::mod_params::{PacketParams, RadioError};
-use super::mod_traits::RadioKind;
+use super::mod_traits::{IrqState, RadioKind, TargetIrqState};
 use super::{DelayNs, LoRa};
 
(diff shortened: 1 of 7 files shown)


lthiery:
`Figure out issue with RX2 not working`
It looks like the unit tests haven't caught this issue. What are you seeing?

Also, it looks like the class C tests are working again and you can delete the noise test/feature IMO.

PR #201, opened 2024-01-10
 Update embedded hal to 1.0.0

Ragarnoy:
This branch is based on lthierry's ehal branch
I'm guessing this is correct, feel free to let me know if not

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,17 @@
+[net]
+git-fetch-with-cli = true
+
+[patch.crates-io]
+embedded-hal = { git = "https://github.com/Dirbaio/embedded-hal", rev = "0e4a3fd08ab919a0b697368305cd3dd926319ed2"}
+embedded-hal-bus = { git = "https://github.com/Dirbaio/embedded-hal", rev = "0e4a3fd08ab919a0b697368305cd3dd926319ed2"}
+embedded-hal-async = { git = "https://github.com/Dirbaio/embedded-hal", rev = "0e4a3fd08ab919a0b697368305cd3dd926319ed2"}
+embedded-hal-nb = { git = "https://github.com/Dirbaio/embedded-hal", rev = "0e4a3fd08ab919a0b697368305cd3dd926319ed2"}
+
+embassy-time = { git = "https://github.com/embassy-rs/embassy.git", rev = "7b1035fd2aec124d47cc9107e5aae56ad3c1cccd" }
+embassy-sync = { git = "https://github.com/embassy-rs/embassy.git", rev = "7b1035fd2aec124d47cc9107e5aae56ad3c1cccd" }
+embassy-rp = { git = "https://github.com/embassy-rs/embassy.git", rev = "7b1035fd2aec124d47cc9107e5aae56ad3c1cccd" }
+embassy-nrf = { git = "https://github.com/embassy-rs/embassy.git", rev = "7b1035fd2aec124d47cc9107e5aae56ad3c1cccd" }
+embassy-stm32 = { git = "https://github.com/embassy-rs/embassy.git", rev = "7b1035fd2aec124d47cc9107e5aae56ad3c1cccd" }
+
+#[patch.'https://github.com/embassy-rs/stm32-data-generated']
+#stm32-metapac = { path = "/home/dirbaio/embassy/stm32-data/build/stm32-metapac" }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,16 +2,16 @@
 git-fetch-with-cli = true
 
 [patch.crates-io]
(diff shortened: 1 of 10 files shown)


lorawan-device/Cargo.toml
```diff
@@ -19,17 +19,17 @@ rustdoc-args = ["--cfg", "docsrs"]
...
+generic-array = "0.14"
 defmt = { version = "0.3", optional = true }
 fastrand = { version = "2", default-features = false }
 futures = { version = "0.3", default-features = false, optional = true }
 rand_core = { version = "0.6", default-features = false }
 serde = { version = "1", default-features = false, features = ["derive"], optional = true }
 seq-macro = "0.3.5"
 document-features = "0.2.7"
-embassy-time = { version = "0.1.5", optional = true }
+embassy-time = { git = "https://github.com/embassy-rs/embassy", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3", optional = true }
```
Dirbaio, line 32: don't use git in `[dependencies]`, it makes it too easy to end up with multiple incompatible `embassy-time` versions which then fails to link. 

This is why the previous code used `crates.io` versions and then replaced them all to git with a single `[patch.crates-io]`. This ensures all `embassy-time` crates in the tree use the same git rev, so no incompatibiltiies.

lora-phy/Cargo.toml
```diff
@@ -20,9 +20,9 @@ lorawan-device = { path = "../lorawan-device", version = ">=0.11.0", features =
     "async",
 ], optional = true }
 num-traits = { version = "0.2", default-features = false }
-
-embedded-hal = { version = "=1.0.0-rc.2" }
-embedded-hal-async = { version = "=1.0.0-rc.2" }
+embedded-hal = { version = "=1.0.0" }
```
Dirbaio, line 23: no need to use `=` anymore, neer `1.x.x` releases are now guaranteed to be backwards-compatible.

.cargo/config.toml
```diff
@@ -0,0 +1,17 @@
+[net]
+git-fetch-with-cli = true
+
+[patch.crates-io]
+embedded-hal = { git = "https://github.com/Dirbaio/embedded-hal", rev = "d90b59b0f48bba1e65129d9f8d5c702d47794d9f"}
```
Dirbaio, line 5: patching embedded-hal should no longer be needed.

For embassy crates I'd keep them in the individual `Cargo.toml` files, it makes it easier to copy one example project out of the repo because it's self contained, but no strong opinion from me, if you prefer centralizing them that's fine too.

Ragarnoy: I feel like it's easier to have them all in the same place, prevents having redundant patches imo

rust-toolchain.toml
```diff
@@ -1,8 +1,8 @@
 # Before upgrading check that everything is available on all tier1 targets here:
 # https://rust-lang.github.io/rustup-components-history
 [toolchain]
-channel = "beta"
-components = [ "rust-src", "rustfmt", "llvm-tools" ]
+channel = "beta-2023-12-17"
```
Dirbaio, line 4: this can be changed to stable now :D

CBJamo:
Sorry for not catching this sooner. For #187, I was planning on waiting for the embassy releases to crates. I believe we'll be able to avoid patches and git deps entirely when that shakes out.

rust-toolchain.toml
Ragarnoy: whoops

Dirbaio (approved):
:rocket:

.cargo/config.toml
```diff
@@ -0,0 +1,12 @@
...
+git-fetch-with-cli = true
+
+[patch.crates-io]
+embassy-time = { git = "https://github.com/embassy-rs/embassy.git", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3" }
+embassy-sync = { git = "https://github.com/embassy-rs/embassy.git", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3" }
+embassy-rp = { git = "https://github.com/embassy-rs/embassy.git", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3" }
+embassy-nrf = { git = "https://github.com/embassy-rs/embassy.git", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3" }
+embassy-stm32 = { git = "https://github.com/embassy-rs/embassy.git", rev = "0027a76bb61f19fcae1fe588ba3ff62660d7f7e3" }
+
+#[patch.'https://github.com/embassy-rs/stm32-data-generated']
```
lthiery, line 11: delete this dead code?

lthiery:
I've merged this for now. We can remove the embassy patches once there is a new release.

PR #202, opened 2024-01-10
lorawan-encoding: Add module with various packet length values

plaes:
Parts of commit extracted from PR #142, and eventualy dependency for #198.

### DIFF
--- a/lorawan-encoding/CHANGELOG.md
+++ b/lorawan-encoding/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this project adheres to [Semantic Versioning](https://semver.org/).
 
-# Unreleased
+## [v0.7.5] - Unreleased
+
+- Add `packet_length` module containing constants for packet component sizes.
 
 ---
 
--- a/lorawan-encoding/Cargo.toml
+++ b/lorawan-encoding/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "lorawan"
-version = "0.7.4"
+version = "0.7.5"
 edition = "2021"
 authors = ["Ivaylo Petrov <ivajloip@gmail.com>"]
 description = "Crate lorawan provides structures and tools for reading and writing LoRaWAN messages from and to a slice of bytes."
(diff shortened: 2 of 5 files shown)


lthiery:
I think there are some issues with some of these values. I am expecting to see JOIN_ACCEPT_MAX equal to 33 but it comes out to only 28 here. It looks like you are neglecting to warp the JoinRequest with MHDR and MIC at the end.

Here are the references from the LoRaWAN Specification

image

The PHYPayload length is what we will want to pass into the lora_modulation calculator. Note: the CRC configuration is accounted for in `BaseBandModulationParams`, so I think your `UPLINK_CRC` and `UPLINK/DOWNLINK_OVERHEAD` should be removed.

You can see from the above figure, we have:
```
DATA_FRAME = MHDR + MACPayload + MIC;
JOIN_REQUEST = MHDR + JOIN_REQUEST_PAYLOAD + MIC;
JOIN_ACCEPT = MHDR + JOIN_ACCEPT_PAYLOAD + MIC;
```

Also, the next figure makes it easy to get `MHDR` and `MIC` but you have set those properly already:

image

So for JOIN_ACCEPT_MAX, you can look at the JOIN_REQUEST frame, which you extracted properly as well:

image

So I think we will see:
```
// JoinNonce, NetID, DevAddr, DLSettings, RxDelay
JOIN_ACCEPT_PAYLOAD_MIN = 3+3+4+1+1
// Optional CFList
JOIN_ACCEPT_PAYLOAD_MAX = JOIN_ACCEPT_PAYLOAD_MIN + 16
```

So the final values you will have are
```
JOIN_ACCEPT_MIN = MHDR + JOIN_ACCEPT_PAYLOAD_IN + MIC;
JOIN_ACCEPT_MAX = MHDR + JOIN_ACCEPT_PAYLOAD_MAX + MIC;
``

**Comment by lthiery** [comment]
Woops, I totally missed the redeclaration of the OVERHEAD variables
```
    pub const UPLINK_OVERHEAD: u8 = UPLINK_CRC;
    pub const DOWNLINK_OVERHEAD: u8 = 0;

    pub mod mac {
        pub const UPLINK_OVERHEAD: u8 = super::UPLINK_OVERHEAD + MHDR + MIC;
        pub const DOWNLINK_OVERHEAD: u8 = super::DOWNLINK_OVERHEAD + MHDR + MIC;
  ```

plaes:
Let's put this PR on hold for now, I managed to figure out the preamble detection on sx1272, so maybe we don't even need this.

lthiery:
heh I liked using it in more in the encoding code like I did in PR to you
https://github.com/plaes/lora-rs/pull/2

lthiery:
Replace by #206

PR #204, opened 2024-01-12
Implement RX1/RX2 listening using RxTimeout from radio

plaes:
This PR implements preamble-detection using Radio timeout and rips out most of the timer functionality during RX1 and RX2 windows. This probably makes #198 obsolete.

Tested with:
* sx1262
* sx1272

### DIFF
--- a/lora-modulation/src/lib.rs
+++ b/lora-modulation/src/lib.rs
@@ -124,6 +124,10 @@ impl BaseBandModulationParams {
         Self { sf, bw, cr, ldro, t_sym_us }
     }
 
+    pub const fn delay_in_symbols(&self, delay_in_ms: u32) -> u16 {
+        (delay_in_ms * 1000 / self.t_sym_us) as u16
+    }
+
     /// Calculates time on air for a given payload and modulation parameters.
     /// If `preamble` is None, the whole preamble including syncword is excluded from calculation.
     pub const fn time_on_air_us(
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -14,7 +14,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 defmt = { version = "0.3" }
-lora-modulation = { version = ">=0.1.2" }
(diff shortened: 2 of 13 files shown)


lthiery:
I tested this on the stm32wl/lora_lorawan example without success. I need to look more into what is going on. My guess is that we're arriving too early perhaps and the timeout fires too soon? I'm not too sure.

lthiery:
I looked more closely at this today and I think I understand the issue a bit better. I believe this "symbol timeout" may work for EU868 because the on-air-time of the 12 join symbols is much longer. It didn't work for US915, because the on-air-time is much less. 

Because of this intuition, I tried reducing `rx_window_offset` from -50. Joins appear to work again from -25 to -3. Part of the reason for this apparently flexibility is because the preamble may be detected even if you don't catch the whole preamble (I remember reading this in a Semtech app note somewhere).

However, tuning `rx_window_offset` to an arbitrary number isn't a very reliable fix; in my opinion, it should work more reliably (albeit less efficiently) the bigger the value gets. In other words, the intention of `rx_window_offset` is to provide some fudge factor; the time before the window where you want to start arming receive. Not only does this give you time to execute the SPI transactions to configure the radio, but it also accounts for the non-deterministic delays between the interrupts and their handling (at least with a cooperative scheduler like embassy). That is to say, I don't have good guarantees from the moment the "send complete" interrupt fires and when I execute the `Timer::reset` which becomes the reference for when I schedule the `Radio::rx_setup`. On top of that, we don't have strong guarantees for when `Timer::at(ms).await` will return, as another task could be executing as we wait for our turn.

Timer Inaccuracy

One simple fix that I think we should consider is to include the `rx_window_offset` in the preamble timeout. But it's difficult to do at the moment because our timeout is symbol based at the moment. Was there a reason to not use the time-based timeout of the SX12xx?

plaes:
Thanks @lthiery, I have pulled in your changes and then did some rebasing and reworking, so commits follow somewhat logical order: https://github.com/plaes/lora-rs/tree/preamble-rxtimeout-rebase

PR #206, opened 2024-01-14
Packet Lengths

lthiery:


lthiery:
Probably want to change the type to usize given that they are mostly used for array indexing

plaes:
> Probably want to change the type to usize given that they are mostly used for array indexing

Yeah, that would get rid of most `.into()` and `as usize`.

PR #211, opened 2024-01-18
lora_phy::sx126x: symbol calc bug fix

lthiery:
It seems a bit shift right was lost in translation from [here](https://github.com/Lora-net/sx126x_driver/blob/master/src/sx126x.c#L807C1-L810C11).


plaes (approved):

plaes:
Confirmed working now: RxTimeout for EU868: ~460ms:
```
6.526794 process_irq loop entered
6.987518 RxTxTimeout in radio mode Receive
```

PR #212, opened 2024-01-18
lorawan-device: remove feature flag for async

lthiery:
Now that async does not require nightly, I don't think there is a reason to hide it behind a feature flag.


lulf (approved):
Agreed!

plaes (approved):

PR #213, opened 2024-01-18
lorawan_device: logs

lthiery:


lulf (approved):
The current approach looks good.

PR #217, opened 2024-01-22
lorawan-device: allow multiple rxc frames during rxc window

lthiery:
In theory, multiple RXC frames may be received between windows. This fixes the issue (and the related test) while also refactoring the RXC code to be more self-contained under `fn between_windows`.


lorawan-device/src/async_device/mod.rs
```diff
@@ -317,11 +346,17 @@ where
         let timeout_fut = self.timer.at(duration.into());
         pin_mut!(timeout_fut);
         let mut maybe_timeout_fut = Some(timeout_fut);
+        //
```
plaes, line 349: Drop it or add something here?

plaes (approved):

PR #219, opened 2024-01-23
lora-phy: sx127x: fix snr and rssi calculation

fe1es:
Fix snr and rssi calculation according to the [datasheet](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/2R0000001Rbr/6EfVZUorrpoKFfvaF_Fkpgp5kzjiNyiAbqcpqh9qSjE) and [an implementation by Semtech](https://os.mbed.com/teams/Semtech/code/SX1276Lib//file/737e89f8d4f3/sx1276/sx1276.cpp/) in cpp:

* Casts packet_snr as i8 before as i16 to maintain the sign then divides it by 4.
* Takes snr and frf into account to calc rssi.

### DIFF
--- a/lora-phy/src/sx1276_7_8_9/mod.rs
+++ b/lora-phy/src/sx1276_7_8_9/mod.rs
@@ -24,9 +24,9 @@ const SX127X_MIN_LORA_SYMB_NUM_TIMEOUT: u16 = 4;
 const SX127X_MAX_LORA_SYMB_NUM_TIMEOUT: u16 = 1023;
 
 // Constant values need to compute the RSSI value
-const RSSI_OFFSET_LF: i16 = -164;
-const RSSI_OFFSET_HF: i16 = -157;
-const RF_MID_BAND_THRESH: u32 = 525_000_000;
+const SX1276_RSSI_OFFSET_LF: i16 = -164;
+const SX1276_RSSI_OFFSET_HF: i16 = -157;
+const SX1276_RF_MID_BAND_THRESH: u32 = 525_000_000;
 
 /// Supported SX127x chip variants
 #[derive(Clone, Copy)]
@@ -565,10 +565,10 @@ where
                 (frf as f64 * FREQUENCY_SYNTHESIZER_STEP) as u32
             };
 
-            let rssi_offset = if frequency_in_hz > RF_MID_BAND_THRESH {
-                RSSI_OFFSET_HF
+            let rssi_offset = if frequency_in_hz > SX1276_RF_MID_BAND_THRESH {
+                SX1276_RSSI_OFFSET_HF
             } else {
-                RSSI_OFFSET_LF
+                SX1276_RSSI_OFFSET_LF
             };
 
             if snr >= 0 {
--- a/lora-phy/src/sx1276_7_8_9/mod.rs
+++ b/lora-phy/src/sx1276_7_8_9/mod.rs
@@ -23,6 +23,11 @@ const FREQUENCY_SYNTHESIZER_STEP: f64 = 61.03515625; // FXOSC (32 MHz) * 1000000
 const SX127X_MIN_LORA_SYMB_NUM_TIMEOUT: u16 = 4;
 const SX127X_MAX_LORA_SYMB_NUM_TIMEOUT: u16 = 1023;
 
+// Constant values need to compute the RSSI value
+const RSSI_OFFSET_LF: i16 = -164;
+const RSSI_OFFSET_HF: i16 = -157;
+const RF_MID_BAND_THRESH: u32 = 525_000_000;
+
 /// Supported SX127x chip variants
 #[derive(Clone, Copy)]
 pub enum Sx127xVariant {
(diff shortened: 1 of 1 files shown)


lora-phy/src/sx1276_7_8_9/mod.rs
```diff
@@ -23,6 +23,11 @@ const FREQUENCY_SYNTHESIZER_STEP: f64 = 61.03515625; // FXOSC (32 MHz) * 1000000
 const SX127X_MIN_LORA_SYMB_NUM_TIMEOUT: u16 = 4;
 const SX127X_MAX_LORA_SYMB_NUM_TIMEOUT: u16 = 1023;
 
+// Constant values need to compute the RSSI value
+const RSSI_OFFSET_LF: i16 = -164;
```
plaes, line 27: Could you update the names of these constants to specify that these are for sx1276.
As this driver supports also sx1272, we'll need to fix RSSI/SNR calculation eventually for it as well.
(sx1272 supports only single range and it's RSSI_OFFSET value is -139).

lora-phy/src/sx1276_7_8_9/mod.rs
```diff
@@ -544,10 +549,35 @@ where
     }
 
     async fn get_rx_packet_status(&mut self) -> Result<PacketStatus, RadioError> {
-        let rssi_raw = self.read_register(Register::RegPktRssiValue).await?;
-        let rssi = (rssi_raw as i16) - 157i16; // or -164 for low frequency port ???
-        let snr_raw = self.read_register(Register::RegPktRssiValue).await?;
-        let snr = snr_raw as i16;
+        let snr = {
```
plaes, line 552: Please include comment:
`// TODO: Fix for sx1272`

plaes (changes requested):

lora-phy/src/sx1276_7_8_9/mod.rs
fe1es: Thanks for the reviews.
Added a prefix to the constants.

lora-phy/src/sx1276_7_8_9/mod.rs
fe1es: Added the comment.

fe1es:
@plaes I tried to fix it for sx1272. Could you review it again? Though the [datasheet](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/440000001NCE/v_VBhk1IolDgxwwnOpcS_vTFxPfSEPQbuneK3mWsXlU) mentions LnaBoost, not sure what to do when it’s Off. It’s also not referenced in an [implementation](https://os.mbed.com/teams/Semtech/code/SX1272Lib//file/b988b60083a1/sx1272/sx1272.cpp/) by Semtech, so I ignored it.

plaes (approved):
> @plaes I tried to fix it for sx1272. Could you review it again? Though the [datasheet](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/440000001NCE/v_VBhk1IolDgxwwnOpcS_vTFxPfSEPQbuneK3mWsXlU) mentions LnaBoost, not sure what to do when it’s Off. It’s also not referenced in an [implementation](https://os.mbed.com/teams/Semtech/code/SX1272Lib//file/b988b60083a1/sx1272/sx1272.cpp/) by Semtech, so I ignored it.

Thanks @fe1es, you beat me to adding sx1272 support. I actually got a mild headache when trying to understand what's happening in relevant section in sx1272 (section 5.5.5) and sx1276 manuals (section 3.5.5) as formulas there use both PacketRssi (`RegPktRssiValue`), and Rssi (`RegRssiValue`) ;)

What their SDK seems to be doing at least for sx1272 in the SNR >=0 case is adding 1/16th of packet_rssi:
`rssi = offset + packet_rssi + (packet_rssi >> 4)`

So instead of `rssi/15` they add `rssi/16` so it's simpler to calculate using bitwise operators:
`rss + rssi/16 => rssi (1 + 1/16) => rss * 17/16`

lora-phy/src/sx1276_7_8_9/mod.rs
```diff
@@ -544,10 +550,40 @@ where
...
+
+                    if frequency_in_hz > SX1276_RF_MID_BAND_THRESH {
+                        SX1276_RSSI_OFFSET_HF
+                    } else {
+                        SX1276_RSSI_OFFSET_LF
+                    }
+                }
+            };
+
+            if snr >= 0 {
```
plaes, line 580: Both sx1272 and sx1276 SDK seem to do following:
```
if snr >= 0 {
    rssi_offset + packet_rssi + (packet_rssi >> 4)
} else {
    rssi_offset + packet_rssi + (packet_rssi >> 4) + snr
}
```

fe1es: Yes, I knew they shift 4 bits while the datasheets say that when SNR >= 0 it's packet_rssi * 16/15 for both sx1272 and sx1276. With the bit shift, It's equals to packet_rssi * 17/16 and I think the difference is negligible, but they add (packet_rssi >> 4) even when SNR < 0 which might be wrong I suppose? Should we follow the SDKs anyway?

plaes: I don't know, to be honest. Current implementation is good enough, and as "perfect is the enemy of the good", we can just start using this.

Thank you again :100:

PR #220, opened 2024-02-08
lora-phy: rename a few things

lthiery:
As I've been poking around the lora-phy codebase, a few names were bugging me. Let me know your thoughts! Naming things is hard.


plaes (approved):
Thanks.

One way to improve this for users would be keeping old names alive via deprecated re-exports, but this feature seems to be still broken https://github.com/rust-lang/rust/issues/84584

CBJamo (approved):
Looks good to me.

PR #222, opened 2024-02-11
lora-phy: refactor IRQ handling and expose more functions

IhsenBouallegue:
Radio Kind was doing more than it should while processing the IRQs like polling and awaiting. It should be a straightforward function that interprets the flags on the radio. Looping/polling/awaiting should be done in the LoRa abstraction as helper functions. This gives more flexibility to anyone using the library to drive the chip however they need.

This also allows us to fix a bug with class c mode where the async branches are not 100% correct and may hang/not detect packages when switching fast between operations. In my use case, I transmit and receive a few times per second. And I verified this works with the SX1272.
new behaviour:
- put chip in rx mode.
  - if we need to transmit, we switch modes to transmit and go back to rx.
  - if we detect an IRQ, we check the IRQ and make sure it's TargetIrqState::Done and get the packet
```rust
loop {
    prepare_for_rx(&mut lora, &mdltn_params, &rx_pkt_params).await;
    tokio::select! {
        // Transmit packets received through channel
        Some(packet) = self.recv_channel.recv() => {
            prepare_for_tx(&mut lora, &mdltn_params).await;
            lora_trans(&mut lora, &packet, &mdltn_params, &mut tx_pkt_params).await;
        }
        // Receive packets from LoRa
        Ok(_) = lora.wait_for_irq() => {
            if let Ok(Some(TargetIrqState::Done)) = lora.process_irq_event(TargetIrqState::Done).await {
                let mut receiving_buffer = [00u8; 255];
                let (received_len, rx_pkt_status) = lora.rx(&rx_pkt_params, &mut receiving_buffer)
                // do something with the received packet
            }
        }
    }
}
```
> Note that I am using tokio but this be the same for embassy

- [x] refactor sx126x
- [x] add back polling next to awaiting in `wait_for_irq` (this is a bit trickier now, but do we even need polling?) -> Polling removed 
- [x] `rx_until_state` should have `wait_for_irq()` at the start to keep old behavior the same, but that breaks my example as I wait for the irq outside the `rx operation`. -> added `process_rx_irq` and refactored methods.
- [x] refactor lorawan class c to detect irq and act on it -> not sure how to do that as I am not familiar with lorawan but it should follow the tokio example above.

> [!NOTE]
> This change doesn't (shouldn't) break the current behaviour. It is mostly refactoring and exposing some internal functions to allow for "class c"-like operation outside of lorawan.


plaes:
Thanks for looking into this. We have had a discussion about polling mode before, and general consensus was that this can be ripped out. So maybe make initial PR to rip this one out first (I have also a WIP unpushed branch for this as well).

About `rx_until_state`: this was actually introduced recently when we started working on the preamble detection support using async timers. But during this process the whole lorawan-device library went through an almost full refactoring causing the preamble branch to "bitrot", therefore it was split up into multiple separate pull requests (lora-phy and lora-rs were then also different git repositories)  and the final preamble flow was totally reworked to use chip timers instead.

lora-phy/src/sx127x/mod.rs
```diff
@@ -713,15 +713,14 @@ where
...
                 if (irq_flags & IrqMask::CADDone.value()) == IrqMask::CADDone.value() {
                     debug!("CADDone in radio mode {}", radio_mode);
-                    // TODO: don't like how we mutate the cad_activity_detected parameter
-                    if let Some(cad_detected_ref) = cad_activity_detected {
-                        // Check if the CAD (Channel Activity Detection) Activity Detected flag is set in irq_flags
-                        let is_cad_activity_detected = (irq_flags & IrqMask::CADActivityDetected.value()) != 0;
-
-                        // Update the referenced value to reflect whether CAD activity was detected
-                        *cad_detected_ref = is_cad_activity_detected;
+                    if (irq_flags & IrqMask::CADDone.value()) == IrqMask::CADDone.value() {
```
plaes, line 716: Could you add back comment about dropping the `cad_activity_detected` argument..

IhsenBouallegue:
@plaes Glad I can help!
I am not using the timeout anymore, I can just rip it completely. So that won't be a problem to also do here. But maybe so things are clearer, I can open a PR directly after this. Let me know what works best!

Concerning the `rx_until_state`, that makes sense because I found no use for it in the library.
So I will refactor this to something along the lines of:
`rx` will call `wait_for_irq` and `process_rx_irq` 
`process_rx_irq` will just call `process_irq_event` but for receiving.

That way in class c we can directly use `wait_for_irq` and `process_rx_irq`. And for everyone that just needs normal `rx`, it won't change.

plaes:
Although I like new code, it seems to be breaking existing lorawan example during the Rx1/Rx2 windows when preamble is received:
1. `lorawan_radio`: rx_single calls lora-phy's rx() which sets up irq wait and then calls match on process_rx_irq(..., TargetIrqState::Done).
2. `process_rx_irq` calls match on process_irq_event with TargetIrqState::Done again.
3. `process_irq_event` returns None because Preamble IRQ is ignored, which then bubbles up as `Err(RadioError::Irq)`  in `process_rx_irq`.

IhsenBouallegue:
@plaes merged your fix and tested my p2p example with the sx1276. It works. I might test with the sx1262, but I have to get that working first, so it will take time.

lucasgranberg:
This works with my lorawan code on stm32wle

plaes (approved):
Tested with sx1261

(Unfortunately device with sx1272 is currently disconnected, but it worked last time I when I developed the fix for this branch and actually nothing has changed so far). 

Unfortunately branch cannot be merged due to conflicts.

plaes:
I have a cleaned up PR that supersedes this one at #231

plaes:
Made obsolete by #231

PR #223, opened 2024-02-12
Drop IRQ polling from process_irq and instead rely on irq pins

plaes:
As far as we understand, there's no real user for irq polling via delay, therefore drop this and simplify the IRQ handling a bit as a result.

Also drops bunch of public but unused Error kinds.


CBJamo:
I like this change. In general we should try to remove any and all polling we can.

Possibly this should be in another PR, but we should look into removing the DLY generic entirely. The two remaining uses can probably be avoided. The use in the sleep function can probably just be dropped (it's already unused on the sx127x). The reset function is a bit trickier, but we might be able to communicate with the device somehow and detect when the reset is complete.

plaes:
> I like this change. In general we should try to remove any and all polling we can.
> 
> Possibly this should be in another PR, but we should look into removing the DLY generic entirely. The two remaining uses can probably be avoided. The use in the sleep function can probably just be dropped (it's already unused on the sx127x). The reset function is a bit trickier, but we might be able to communicate with the device somehow and detect when the reset is complete.

Yeah, I would rather not touch it at the moment and wait a bit until the other WIP improvements (chip trait instead of enum variant and irq refactor) have landed.

CBJamo (approved):
Sounds reasonable. These changes look good to me.

PR #232, opened 2024-03-21
Examples: Add support for overriding lorawan keys during build via LORA_APPKEY, LORA_DEVEUI.. values

plaes:
Implement lorawan key customization via environment variables (for nrf52840):
`LORA_APPEUI= LORA_APPKEY= LORA_DEVEUI=1122334455667788 cargo build --release`


lucasgranberg (dismissed):

lthiery:
What about providing this as a util in the lorawan module? You could call the util from build.rs.

lthiery:
Come to think of it, providing this as a util in the lorawan module will be difficult because this (and build.rs) can depend on std but lorawan can't. 

We could create lorawan-util crates that has std... but I think this is also fine the way it is.

examples/nrf52840/build.rs
```diff
@@ -10,14 +10,77 @@
 
 use std::env;
 use std::fs::File;
-use std::io::Write;
+use std::io::{BufWriter, Write};
 use std::path::PathBuf;
 
+fn hex_to_bytes(s: &str) -> Option<Vec<u8>> {
```
lthiery, line 16: hex::encode?

examples/nrf52840/build.rs
```diff
@@ -10,14 +10,77 @@
...
+        write!(
+            &mut file,
+            "{}",
+            format_args!(
+                "\
+            // Generated by build.rs\n\
+            const DEVEUI: Option<[u8; 8]> = {};\n\
+            const APPEUI: Option<[u8; 8]> = {};\n\
+            const APPKEY: Option<[u8; 16]> = {};\n",
+                parse_lorawan_id(option_env!("LORA_DEVEUI"), "LORA_DEVEUI", 8).unwrap_or("None".to_string()),
```
lthiery, line 74: perhaps implement `DevEui/AppEui/AppKey::from_str`? Feature-gated behind std being enabled in lorawan-encoding.

lthiery:
Actually, I was wrong above. `build-dependencies` could have std enabled for `lorawan-device` which could provide this utility without having std enabled in the project itself. Then that could enable the `from_str` from `lorawan-encoding` (see review)?

lthiery (approved):
Some non-blocking suggestions.

lthiery:
I also see that we hand-rolled the [hex encoding here](https://github.com/lora-rs/lora-rs/blob/main/lorawan-encoding/src/extra/to_string.rs) in lorawan-encoding. I've corrected there [here](https://github.com/lora-rs/lora-rs/commit/67bf7eae2b4322ef59e6b136806fa7fb97176dec).

lthiery:
I went ahead and just made the PR for lorawan-encoding: https://github.com/lora-rs/lora-rs/pull/234

examples/nrf52840/build.rs
plaes: Yeah, I initially didn't want to include extra dependency for just a single semi-trivial function (`left-pad` :dagger: )..

lucasgranberg:
just for consideration the hex decoding can be done at compile time without custom build scripts by using the hex-literal crate. https://crates.io/crates/hex-literal

plaes:
So, I basically ran into a headache and current version is good enough.
I can easily use the `build.rs` to parse the DevEui and friends, but in the end I would need to somehow "tell" rust to use this value in some way, which means I would have to either convert it back to array syntax or find something better.
This is how it works for now:
1. In generated file I have: `const DEVEUI: Option<[u8; 8]> = None;`
2. In lorawan example code  I need to match the types so I can override: `deveui: DevEui::from(DEVEUI.unwrap_or(DEFAULT_DEVEUI)),`

(Also, `from_str()` requires pulling in `use core::str::FromStr;` and its a lot easier to type out zeroes as array than into hex string..)

PR #233, opened 2024-03-25
smtc-modem-cores

lthiery:
After seeing [Semtech's new reference drivers](https://github.com/Lora-net/SWL2001), I thought it would be interesting to start using them in testing the lora-phy implementation. Full chip support exists in these reference drivers, so it would be further useful for developing sx128x or lr1110 implementations.

Furthermore, a no-async device implementation could be based on this bindgen package.

Per review feedback, the bindings crates live in their own repository: https://github.com/lora-rs/smtc-modem-cores. This PR is just the lora-phy test integration — the bindings come in as a git dev-dependency, and the tests compare the SPI byte stream our sx126x driver produces against the reference driver's, starting with `set_sleep` (cold and warm start).

On the C-callback wiring: an earlier iteration used a `concrete_export!` macro that made the user define the `#[no_mangle] sx126x_hal_write` symbol for their concrete SPI type, which was awkward and limited each binary to a single SPI type (duplicate symbol otherwise). The bindings crate now owns that symbol itself and dispatches through a monomorphized fn pointer stored at the head of the `repr(C)` context — `Context::new(spi)` is the whole user-facing API, and multiple SPI types can coexist in one binary. Extending to `sx126x_hal_read`/`reset`/`wakeup` is just more fields in that table.

**Update: lr1110 coverage.** The bindings repo now also builds SWL2001's lr11xx radio core (its HAL differs — two-transaction reads plus a direct read, so the vtable carries three entry points), and this branch adds 46 comparison tests for the lr1110 driver mirroring the sx126x suite: system commands, the LoRa TX/RX/CAD paths, TX power on all three PAs, regmem, GFSK configuration and LR-FHSS framing.

The comparison immediately caught real encoding bugs in the lr1110 driver, fixed here in their own commit: the SX126x sleep-config bit layout (every sleep was cold, losing retention), the SX126x mantissa/exponent symbol timeout (8 symbols became 32), a stray offset byte in WriteBuffer8 (would have been transmitted as the first payload byte), a zero length field in ReadBuffer8, SX126x-style bitrate/deviation conversions in GFSK modulation params (the lr11xx takes raw bps/Hz), a GFSK packet-status decode reading payload length from an RSSI byte, missing AutoTxRx fields, and the high-ACP workaround the reference applies on every SetTx/SetRx/SetCad.

**Update: sx127x coverage.** The register-based sx127x needed a different comparison approach: the two drivers legitimately factor their register traffic differently (SWL2001 burst-writes register runs and keeps shadow copies; ours does single-register read-modify-write over SPI), so the sx127x fixture is a small register file both drivers execute against, and equality is the chip-visible outcome — final register state plus the FIFO byte stream. 16 tests cover the LoRa config surface, TX/CAD flows, RX reads and decoded-value parity.

This round caught five bugs in our sx127x driver, fixed in their own commit: the PLL-step shortcut truncated the low 8 bits so fractional-MHz channels (868.1, 903.9 MHz) sat up to ~15 kHz off carrier; PA_BOOST requests above 17 dBm used the PaDac-off power formula and overflowed the 4-bit OutputPower field; RFO requests at or below 0 dBm cast a negative OutputPower to u8 and corrupted RegPaConfig; the RegDetectionOptimize RMW mask dropped the AutomaticIFOn bit; and packet RSSI skipped the 16/15 linearization in the negative-SNR branch. It also surfaced a reference-driver bug worth knowing about: SWL2001's `sx127x_set_ocp_value` masks the trim with the inverted-form `OCP_TRIM_MASK`, so it always programs 45 mA regardless of the value passed.

With sx126x, lr1110 and sx127x covered, every lora-phy driver with a SWL2001 counterpart now has a comparison suite.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,10 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+      # Needed to build the smtc-modem-cores bindings used by lora-phy tests
+      - name: Setup apt packages
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends cmake llvm-dev clang libclang-dev
+
       - name: Setup Cache
         uses: Swatinem/rust-cache@v2
         with:
@@ -87,6 +91,10 @@ jobs:
         uses: styfle/cancel-workflow-action@0.12.1
         with:
           access_token: ${{ github.token }}
+
+      # Needed to build the smtc-modem-cores bindings used by lora-phy tests
+      - name: Setup apt packages
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends cmake llvm-dev clang libclang-dev
 
       - name: Setup Cache
         uses: Swatinem/rust-cache@v2
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "70c6e87" }
+smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "42d8559" }
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "1949152" }
+smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "f2b684b" }
 tokio = { version = "1", features = ["rt", "macros"] }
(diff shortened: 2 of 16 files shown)


.github/workflows/rust.yml
```diff
@@ -19,11 +19,17 @@ jobs:
...
+      - name: setup submodules
+        run: git submodule update --init --recursive
+
       - name: Cancel previous runs
         uses: styfle/cancel-workflow-action@0.5.0
         with:
           access_token: ${{ github.token }}
 
+      - name: Setup apt packages
+        run: sudo apt-get update && sudo apt-get install -y cmake llvm-dev clang libclang-dev
```
plaes, line 31: `apt-get install -y --no-install-recommends`

smtc-modem-cores/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "smtc-modem-cores"
+version = "0.1.3"
+authors = ["Louis Thiery <thiery.louis@gmail.com>"]
+edition = "2018"
```
plaes, line 5: Use 2021 edition? (Here and other `Cargo.toml` files as well.)

.github/workflows/rust.yml
lthiery: didn't know about this one! thanks

smtc-modem-cores/Cargo.toml
lthiery: copy-pasted an old project to get my started! it shows :smile:

plaes:
I would rather see smtc-modem-cores (together with -sys) shoul in a separate repository.

lthiery:
Rebased on main and restructured: the bindings crates now live in https://github.com/lora-rs/smtc-modem-cores as suggested, so this PR is down to the lora-phy test integration — the smtc crates come in as a git dev-dependency. The set_sleep comparison tests pass against SWL2001 v4.8.0.

lthiery:
Closing in favor of #445, which carries the same branch history restructured under a clearer description — the scope grew well past the original sx126x set_sleep comparison (all three chips are now covered, with a per-module summary of the 14 driver bugs the comparison caught).

PR #234, opened 2024-03-26
Hex Strings

lthiery:


plaes (approved):

plaes:
You may also want to revise copyright strings...

lucasgranberg (approved):

lthiery:
On second though, I realized feature-gating from_str was not terribly useful, so I've reorganized things so that from_str is always available.

lorawan-encoding/src/string.rs
```diff
@@ -0,0 +1,168 @@
...
+                hex::decode_to_slice(s.as_bytes(), &mut res)?;
+                Ok(Self::from(res))
+            }
+        }
+
+        #[cfg(feature = "with-to-string")]
+        impl std::string::ToString for $type {
+            fn to_string(&self) -> std::string::String {
+                let mut res = std::string::String::with_capacity($size * 2);
+                res.extend(std::iter::repeat('-').take($size * 2));
```
plaes, line 27: Why `-`?
Is there a test with under- and oversized hex string?

lthiery: We never assert size, but the `-` ensure an unwritten char would be obvious.

lthiery: Concerning missized strings, good point! I'll work on that now

lthiery: The hex library handles them nicely it seems

PR #237, opened 2024-04-02
make prepare_for_rx behave as expected

lucasgranberg:
prepare_for_rx should do everything that can be done ahead of time before actually starting the rx. This is important for tighter timing ex. for lorawan. rx_continuous is also removed as a state as it can be derived from listen_mode of from the context.


lucasgranberg:
These changes do not play well with the lorawan code. Back to the drawing board..

PR #245, opened 2024-04-04
 modify prepare_for_x() and x() for better timings (rx,tx,cad)

lucasgranberg:
for rx,tx and cad operations
move as much as possible to the prepare step without actually running do_x()

This helps with tighter timings for lorawan


plaes (commented):
Tested with LoRaWAN example with sx1261 / nrf52840 combo and sx1272 radio.

Could you rebase the branch to get rid of merge commits? I can help you with that if needed.

plaes (commented):
Tested again, and works with my radios (sx1272 and sx1261 w/ nrf52840).

One issue with this PR is that each commit should keep a repository in a state where everything compiles (and better if it works as well) so repository is bisectable.
In this case, 2nd commit actually breaks the build for examples thus last two commits should be merged into one.

lucasgranberg:
Yeah idk why I tried to divvy it up between commits. Should have just done one commit. Does it matter if we squash and merge? The library is broken anyway until we get the irq stuff fixed again.

plaes (approved):

PR #247, opened 2024-04-05
allow process_irq_event to continue on RFU

lucasgranberg:
Stm32wl returns a lot of RFU events. These break both rx an tx.


plaes (approved):
Tested with non-stm radios!

PR #221, opened 2024-02-08
lora-phy: refactor sx127x to have chip trait instead of enum

lthiery:
Example of how to factor out the enum in favor of a trait. Sx126x would need to be done next.

PR #248, opened 2024-04-10
lora-phy: refactor sx127x to have chip trait instead of enum

plaes:
This is a rebased and conflict-resolved branch that supersedes #221


lucasgranberg (approved):
This is a good change, lets do the same for sx126x. This should also let people use weird hardware. thanks @lthiery and @plaes

plaes:
Also, this PR reduces binary size, for example the `stm32l0` with LTO enabled:
* `lora_lorawan` 94.4KiB -> 93.1KiB
* `lora_p2p_send` 45.3KiB -> 44.3KiB
* `lora_p2p_receive` 52.1KiB -> 52.0KiB

PR #249, opened 2024-04-10
make sx126x variants trait based

lucasgranberg:
Do the same for sx126x as https://github.com/lora-rs/lora-rs/pull/248 did for sx127x. This PR should only refactor code but not change how it works.

These changes are made to make the code base more uniform between chips but also for allowing better control over HP/LP modes especially for stm32wl.

### DIFF
--- a/lora-phy/src/sx126x/mod.rs
+++ b/lora-phy/src/sx126x/mod.rs
@@ -9,6 +9,8 @@ use radio_kind_params::*;
 use crate::mod_params::*;
 use crate::mod_traits::IrqState;
 use crate::{InterfaceVariant, RadioKind, SpiInterface};
+mod variant;
+pub use variant::*;
 
 // Syncwords for public and private networks
 const LORA_MAC_PUBLIC_SYNCWORD: u16 = 0x3444; // corresponds to sx127x 0x34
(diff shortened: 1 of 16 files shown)


lucasgranberg:
```rs
/// Minew modules only use HighPowePA
pub struct Minew;
impl Sx126xVariant for Minew {
    fn get_device_sel(&self, _output_power: i32) -> DeviceSel {
        DeviceSel::HighPowerPA
    }
}
```
With this change something like this can be done.

lora-phy/src/lib.rs
```diff
@@ -199,7 +199,7 @@ where
                 self.wait_for_irq().await?;
                 match self.radio_kind.process_irq_event(self.radio_mode, None, true).await {
                     Ok(Some(IrqState::Done | IrqState::PreambleReceived)) => {
-                        self.radio_mode = RadioMode::Standby;
+                        //self.radio_mode = RadioMode::Standby;
```
plaes, line 202: Where do we manually switch `self.radio_mode = RadioMode::Standby;` after tx operation?

lucasgranberg: both changes sneaked trough, the Standby after tx is another issue. Right now prepare_for_tx puts in transmit mode and ideally it should be possible to transmit multiple times with one setup but it is not possible as tx() goes to standby. Ill fix my PR later. Thanks for catching

lora-phy/src/sx126x/mod.rs
```diff
@@ -3,13 +3,21 @@ mod radio_kind_params;
 use defmt::debug;
 use embedded_hal_async::delay::DelayNs;
 use embedded_hal_async::spi::*;
+pub use radio_kind_params::Sx126xVariant;
```
plaes, line 6: Put these two on single line?

plaes (commented):
Bunch of nitpicks regarding docstrings, but could you instead of merge just rebase and force-push your branch?

lora-phy/src/sx126x/mod.rs
lucasgranberg: I moved all the variant thing to a separate module. It didn't really belong in radio_kind_params anyway

plaes (approved):
Let's push this out (seems to be working with my only sx126x device).

PR #250, opened 2024-04-16
lora-phy: call wait_for_irq before process_irq_event in LoRa::cad

fe1es:
In LoRa::cad, it doesn't wait for irq before processing it and panics on `unreachable!()`. This fixes the issue.

Tested on sx1276.


lucasgranberg (approved):
Good catch, maybe this could be changed in the future so that process_irq_event would handle it. Should probably be called something else. Maybe there is a reason that I don't know for separating these.

PR #251, opened 2024-04-17
lora-phy: add get_rssi() to RadioKind trait

fe1es:
I need to do carrier sensing using the current RSSI value. The CAD isn't enough because I want to detect waves besides LoRa. So I added `get_rssi()` to the `RadioKind` trait and exposed it. I also added `RadioMode::Listen` in which the radio keeps listening to a given frequency so that the `RegRssiValue` register is updated continuously. I'm not sure if it's ok to add a new RadioMode though.

I don't own sx126x chips, so I implemented `get_rssi()` only for sx127x and tested on sx1276. I'd appreciate it if someone could implement it for sx126x.

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -275,6 +275,11 @@ where
         }
     }
 
+    /// Get the current rssi
+    pub async fn get_rssi(&mut self) -> Result<i16, RadioError> {
+        self.radio_kind.get_rssi().await
+    }
+
     /// Prepare the Semtech chip for a channel activity detection operation
     pub async fn prepare_for_cad(&mut self, mdltn_params: &ModulationParams) -> Result<(), RadioError> {
         self.prepare_modem(mdltn_params).await?;
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -170,7 +170,7 @@ where
         output_power: i32,
         buffer: &[u8],
     ) -> Result<(), RadioError> {
-        self.prepare_modem(mdltn_params).await?;
+        self.prepare_modem(mdltn_params.frequency_in_hz).await?;
 
         self.radio_kind.set_modulation_params(mdltn_params).await?;
         self.radio_kind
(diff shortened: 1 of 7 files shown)


lucasgranberg:
Could you write an example on how you want to use this code?

fe1es:
What I want is basically the same as [this function](https://github.com/ARMmbed/mbed-os/blob/7ae592dabe59e9f26a4ce56190af61fb8b46d10d/connectivity/drivers/lora/COMPONENT_SX1276/SX1276_LoRaRadio.cpp#L1018). With the function, you first check if the RSSI is below a given threshold for a certain duration, and if so, you perform tx(). Otherwise, you try again later.

lora-phy/src/lib.rs
```diff
@@ -275,9 +275,32 @@ where
...
+            self.radio_kind.do_rx(RxMode::Continuous).await?;
+            Ok(())
+        } else {
+            Err(RadioError::InvalidRadioMode)
+        }
+    }
+
+    /// Get the current rssi
+    pub async fn get_rssi(&mut self) -> Result<i16, RadioError> {
+        self.radio_kind.get_rssi().await
```
plaes, line 298: Shouldn't we check whether we are in correct mode?

plaes (changes requested):

lora-phy/src/lib.rs
fe1es: I skipped mode checking because I though someone might want to use the function regardless of the current mode, such as checking the last RSSI (not the packet one) after performing rx() is done or timed out. It would be unable to do such things If we restrict it to listen mode only and remove `prepare_listen`, but I personally don't mind either way.

plaes: Ok, makes sense!

plaes (dismissed):

lucasgranberg:
This makes sense to me. I would still like some example code at least for one target on how to use this. It helps us with maintaining the code and it works as documentation too. Also it makes it easier for anyone with the hw to test it.

fe1es:
I added a very simple example for stm32l0 which just keeps logging the current rssi.

lthiery:
Is this PR still a draft or is it ready for final review?

fe1es:
I think this is ready for review.

lthiery:
superseded by #261

PR #259, opened 2024-05-28
lorawan-radio: add the low power feature

pcurt:
This feature allows to set the radio in low power mode when we are using a lorawan device.

The `low_power` is automatically called after a lorawan transmission,
this function is now calling the lora chip sleep function.

Consumption has been measured in sleep mode after a lorawan communication with a SX1262 module and is in the expected range (1µA), the others 6µA comes from others components.

327426836-2c795886-8783-467a-9e88-cf8bc7916b95

### DIFF
--- a/lora-phy/src/lorawan_radio.rs
+++ b/lora-phy/src/lorawan_radio.rs
@@ -156,6 +156,9 @@ where
             Err(Error::NoRxParams)
         }
     }
+    async fn low_power(&mut self) -> Result<(), Self::PhyError> {
+        self.lora.sleep(false).await.map_err(|e| e.into())
+    }
 }
 
 impl RxMode {


lorawan-device/src/async_device/radio.rs
lucasgranberg, line 1: No need to change the trait implementation. It only acts as a default so the impl is not required to do low_power.

lora-phy/src/lorawan_radio.rs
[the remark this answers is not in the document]
lthiery: yea i think that'd be more idiomatic

lorawan-device/src/async_device/radio.rs
lthiery: :100:

lthiery (changes requested):

lora-phy/src/lorawan_radio.rs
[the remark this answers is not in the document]
pcurt: Yes, I agree, thanks for the suggestion

lorawan-device/src/async_device/radio.rs
pcurt: Yes, I have reverted the the modification

lucasgranberg (approved):
Looks good, thanks for the PR and for taking the time to measure this.

plaes (approved):

lthiery (approved):

PR #261, opened 2024-07-02
lora-phy: add get_rssi() to RadioKind trait and include sx126x

### LINKED ISSUE #254: Implement get_rssi() for sx126x
PR #251 currently adds support only for sx127x, though similar functionality for sx126x requires only calling the `GetRssiInst = 0x15` Op and returning the result.

ionspin:
This is an expansion of #251 that adds support for sx1262.

One significant change I had to include was adding modulation parameters to `listen()` method, as it seems to set the bandwith the chip will listen on. Without it getRSSIInst didn't refresh RSSI values on subsequent calls. 

I don't have a 127x based device to test with, so I can't tell if this change to `listen` method will affect the behavior on those device. It would be great if @fe1es could confirm if it works as expected, otherwise I guess we would need to push the implementation of `listen` down to RadioKind?

I've also added an example for nrf52840 based RAK4631 and rebased to current main.

Fixes #254


fe1es:
@ionspin Thanks for adding support for sx1262.
Your code panics because `SpreadingFactor::_5` is unavailable for sx127x, so I had to use something available.
It seems that the bandwidth affects the rssi values, e.g., the values with 500kHz are almost identical to those without the bandwidth, but it gets slightly higher with 125kHz. So I think we could add the bandwidth as a parameter for `listen()` and `set_bandwidth()` on RadioKind.

ionspin:
> @ionspin Thanks for adding support for sx1262. Your code panics because `SpreadingFactor::_5` is unavailable for sx127x, so I had to use something available. It seems that the bandwidth affects the rssi values, e.g., the values with 500kHz are almost identical to those without the bandwidth, but it gets slightly higher with 125kHz. So I think we could add the bandwidth as a parameter for `listen()` and `set_bandwidth()` on RadioKind.

Thanks for taking the time to test this out on 127x, and thanks for the original pull request!

I feel like that's the way to go as well, I'll update the pull request when I get some free time.

ionspin:
I have added bandwidth to the listen function, but I haven't added `set_bandwidth` to `RadioKind`, because as far as I can see setting just the bandwidth without affecting other modulation parameters is not possible on sx126x.

image

And there is no way to retrieve current modulation params, so that they can be modified with just the bandwitdth parameter.

fe1es:
Since we have to pass modulation params for all `prepare_for_xx` methods, the hard-coded spreading factor and coding rate shouldn't affect subsequent operations. So for now, we could choose params that are available on all chips.
By the way, you shouldn't unwrap on `create_modulation_params` as it checks if the bandwidth is available and if the combination of the bandwidth and the frequency is valid too.

ionspin:
> Since we have to pass modulation params for all prepare_for_xx methods, the hard-coded spreading factor and coding rate shouldn't affect subsequent operations. So for now, we could choose params that are available on all chips.

I've checked in existing code, and the datasheets for 1261, 1262, 1272/72, 1276/77/78/79 and all seem to support SF7 and 4/5 coding.

> By the way, you shouldn't unwrap on create_modulation_params as it checks if the bandwidth is available and if the combination of the bandwidth and the frequency is valid too.

Thanks, I'm new to rust, and still haven't internalized best practices (or in this case common sense :) ), so it helps to get some advice.

lthiery:
Nice work on the collaboration @ionspin and @fe1es. Sorry for taking so long to review - I'm not very comfortable with the radio driver code.

This looks good to me. One nit is that I noticed the signature of `prepare_modem` had changed and I was a bit confused how one can prepare a modem with only frequency (I know SF and BW are important!). From the looks of it, `prepare_modem` only sets the frequency and does some other work to "stabilize it" to receive? I wish there was a more descriptive fn name but I don't know what it is. 

Anyway, it's just a nit in my book. If you could rebase this branch we should finally land this IMO unless anyone else more comfortable with the radio drivers wants to chime in.

lthiery (approved):

plaes:
> Nice work on the collaboration @ionspin and @fe1es. Sorry for taking so long to review - I'm not very comfortable with the radio driver code.
> 
> This looks good to me. One nit is that I noticed the signature of `prepare_modem` had changed and I was a bit confused how one can prepare a modem with only frequency (I know SF and BW are important!). From the looks of it, `prepare_modem` only sets the frequency and does some other work to "stabilize it" to receive? I wish there was a more descriptive fn name but I don't know what it is.

We should actually store frequency as part of modem state. This is also done in official semtech modem drivers.

This would also allow us to get rid of extra register access when calculating RSSI for sx127x driver:
https://github.com/lora-rs/lora-rs/blob/969ee2406d957954921d4be6fda69adac976534b/lora-phy/src/sx127x/sx1276.rs#L137-L154

PR #268, opened 2024-07-11
try to fix docs

lthiery:


lora-phy/src/lib.rs
```diff
@@ -1,8 +1,8 @@
 #![no_std]
 #![allow(async_fn_in_trait)]
 #![warn(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![doc = include_str!("../README.md")]
-
```
plaes, line 5: Please keep this newline?

lthiery: right on. i'll fix in the new PR

PR #269, opened 2024-07-11
build docs as part of CI

lthiery:


lorawan-encoding/src/creator.rs
```diff
@@ -22,8 +22,7 @@ use super::default_crypto::DefaultFactory;
 
 use super::keys::Decrypter;
```
plaes, line 23: While at it, can you merge this line into line 15...

plaes (approved):
Looking good...

plaes (approved):

PR #273, opened 2024-07-16
Allow AFIT where needed to suppress warnings

plaes:


lucasgranberg:
Can this be done globally by placing `#![allow(async_fn_in_trait)]
`  in lib.rs of lora-phy and lorawan-device?

plaes:
> Can this be done globally by placing `#![allow(async_fn_in_trait)] ` in lib.rs of lora-phy and lorawan-device?

It can be (and it was like that before), but I would rather be explicit and only tag traits that actually need this.

lucasgranberg (approved):
I think allowing in lib.rs would be better as lorawan-device and lora-phy will be used in single threaded embedded context, at least so far. If not than then actually doing the whole Feature + Send thing could work too.

Anyway I don't feel that strongly about it, lets do it your way.

lthiery:
I don't have a preference on style here. I think this works!

lthiery (approved):

PR #288, opened 2024-09-20
lora-phy: Add API for specifying the sync word

jeholliday:
Hello, I have created this as a PR but I am open to feedback if you see a better way to accomplish this. I would like to allow users of this library to specify their own sync word when using the API. Currently, it is not possible to use any sync word other than the standard LoRaWAN ones. However other networks, such as Meshtastic, may use a different sync word.

As a part of this change, I have also improved the sx162x driver to be able to use the same sync word as the sx172x driver. This change calculates the two byte sync word config from the one byte sync word the same way as RadioLib: https://github.com/jgromes/RadioLib/blob/8068bcc3567350a736af32a5efddd86c69d5e99f/src/modules/SX126x/SX126x.cpp#L892.

I have only tested this with an sx1262 device, but there sure be no impact to functionality with any device. I have not run any of the examples with this change, so I would appreciate if someone could verify that I have not broken any of them. I purposefully did not want to change the existing new function so that the examples should continue to work. I have checked that the tests are successful.

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -38,37 +38,6 @@ const LORAWAN_PUBLIC_SYNCWORD: u8 = 0x34;
 /// Sync word for private LoRaWAN networks
 const LORAWAN_PRIVATE_SYNCWORD: u8 = 0x12;
 
-/// Sync word for Meshtastic network
-const MESHTASTIC_SYNCWORD: u8 = 0x2B;
-
-/// Sync word to differentiate LoRa networks
-#[derive(Clone, Copy, PartialEq, Eq)]
-pub enum NetworkSyncWord {
-    /// Public LoRaWAN network sync word
-    LoRaWanPublic,
-
-    /// Private LoRaWAN network sync word
-    LoRaWanPrivate,
-
-    /// Meshtastic network sync word
-    Meshtastic,
-
-    /// Custom network sync word
-    CustomNetwork(u8),
-}
-
-impl NetworkSyncWord {
-    /// Convert the network sync word to a byte
-    pub fn as_byte(self) -> u8 {
-        match self {
-            NetworkSyncWord::LoRaWanPublic => LORAWAN_PUBLIC_SYNCWORD,
-            NetworkSyncWord::LoRaWanPrivate => LORAWAN_PRIVATE_SYNCWORD,
-            NetworkSyncWord::Meshtastic => MESHTASTIC_SYNCWORD,
-            NetworkSyncWord::CustomNetwork(n) => n,
-        }
-    }
-}
-
 /// Provides the physical layer API to support LoRa chips
 pub struct LoRa<RK, DLY>
 where
@@ -78,7 +47,7 @@ where
     radio_kind: RK,
     delay: DLY,
     radio_mode: RadioMode,
-    sync_word: NetworkSyncWord,
+    sync_word: u8,
     cold_start: bool,
     calibrate_image: bool,
 }
@@ -89,7 +58,7 @@ where
     DLY: DelayNs,
 {
     /// Build and return a new instance of the LoRa physical layer API with a specified sync word
-    pub async fn with_syncword(radio_kind: RK, sync_word: NetworkSyncWord, delay: DLY) -> Result<Self, RadioError> {
+    pub async fn with_syncword(radio_kind: RK, sync_word: u8, delay: DLY) -> Result<Self, RadioError> {
         let mut lora = Self {
             radio_kind,
             delay,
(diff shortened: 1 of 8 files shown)


plaes:
Please add a test to show that sx126x network key is properly converted.

Also, how about using enum?

```rust
pub enum NetworkSyncWord {
  LoRaWanPublic,
  LoRaWanPrivate,
  MeshTastic,
  CustomNetwork(u8),
}

impl LoRaSyncWord {
    // TODO: find a good name...
    pub fn value(self) -> u8 {
        match self {
            NetworkSyncWord::LoRaWanPublic => LORAWAN_PUBLIC_SYNCWORD,
            NetworkSyncWord::LoRaWanPrivate => LORAWAN_PRIVATE_SYNCWORD,
            NetworkSyncWord::MeshTastic => ???,
            NetworkSyncWord::CustomNetwork(n) => n,
        }        
    }
}
```

lucasgranberg:
docs for sx1276 says 
> Note Sync Word values containing 0x00 byte(s) are forbidden
Should we check that?

I prefer using enum like plaes suggested, it makes the API easier to use and it documents itself

jeholliday:
Thanks for the suggestions. I have added the enum and a test.

plaes (approved):

jeholliday:
Will my changes be squashed into one commit when they are merged in?

plaes:
> Will my changes be squashed into one commit when they are merged in?

Which way do you want them?

jeholliday:
> > Will my changes be squashed into one commit when they are merged in?
> 
> Which way do you want them?

I prefer squashing. I just haven't used GitHub for PRs in a while so I'm not sure how to do that.

lthiery:
Sorry, I'm late to the party. I am of the opinion that the radio API+code, as a rule, should never make mention of any specific protocols.

As such, I'd recommend the API simply be: `with_syncword(radio_kind: RK, sync_word: u8, delay: DLY)`. If you want to be kind to the user, include the LoRaWanPublic, LoRaWanPrivate, and Meshtastic values in  the documentation of the API.

If both of you both prefer this, I won't argue further. Just sharing my 2 cents.

jeholliday:
> Sorry, I'm late to the party. I am of the opinion that the radio API+code, as a rule, should never make mention of any specific protocols.
> 
> As such, I'd recommend the API simply be: `with_syncword(radio_kind: RK, sync_word: u8, delay: DLY)`. If you want to be kind to the user, include the LoRaWanPublic, LoRaWanPrivate, and Meshtastic values in the documentation of the API.
> 
> If both of you both prefer this, I won't argue further. Just sharing my 2 cents.

That was also my thinking with my initial implementation of the API, that the LoRa physical layer API shouldn't have any knowledge about the layers above it. However, I will defer to what the maintainers think is best. It's not too late to change it.

I am not sure why this PR is just stuck in approved. Is there something else I need to do?

lucasgranberg:
> I am not sure why this PR is just stuck in approved. Is there something else I need to do?

We just have to agree on the approach and do the merge. I liked the enum approach but it has its flaws. Our "list" will never be complete as we don't know what is being used. Thinking about it now I start to agree with @lthiery. Sorry for the back and forth.

The way `fn new` is set up makes it easy for the user to select between the most common use cases (private/public) anyway so my point about having the enum document the value is kind of moot.

What do you think @plaes?

plaes:
> > I am not sure why this PR is just stuck in approved. Is there something else I need to do?
> 
> We just have to agree on the approach and do the merge. I liked the enum approach but it has its flaws. Our "list" will never be complete as we don't know what is being used. Thinking about it now I start to agree with @lthiery. Sorry for the back and forth.
> 
> The way `fn new` is set up makes it easy for the user to select between the most common use cases (private/public) anyway so my point about having the enum document the value is kind of moot.
> 
> What do you think @plaes?

We sometimes think we have excellent ideas, but this is always based on a certain level of experience we have accumulated by this point of time. That's one of the reasons we were waiting for @lthiery's feedback.
I now also agree that having a public API listing all the current and future LoRa networks is not a good idea...

I'll pull in two first commits as they are and also testcase from 3rd one.

jeholliday:
@plaes Your changes look good, and I have pulled them in.

plaes (approved):
\o/

lora-phy/src/lib.rs
```diff
@@ -66,6 +72,20 @@ where
...
+    /// control an initialized LoRa radio for LoRaWAN public or private network.
+    ///
+    /// In order to configure radio to use non-LoRaWAN networks, use
+    /// [`Self::with_syncword()`] which has `sync_word` argument.
+    pub async fn new(radio_kind: RK, enable_public_network: bool, delay: DLY) -> Result<Self, RadioError> {
+        let sync_word = if enable_public_network {
+            LORAWAN_PUBLIC_SYNCWORD
+        } else {
+            LORAWAN_PRIVATE_SYNCWORD
+        };
```
lthiery, line 85: So I guess this kind of violates our rules about protocol level stuff in the radio-driver, but it's maintained to preserve the API?

lthiery (approved):

lora-phy/src/lib.rs
plaes: Yeah, this is also defined in trait:
https://github.com/lora-rs/lora-rs/blob/5664d38c692da8574adc1dd353de29d307540ab4/lora-phy/src/mod_traits.rs#L36-L37

lthiery:
superseded by #296

Issue #126, opened 2023-09-11
deprecate use of GenericArray


lthiery:
With const generics, it should be possible now. I think we use GenericArray because it impls Default though. I'm not sure what we need to use instead @lulf 

eg: 
```rust
let mut phy: DataPayloadCreator<[u8; 255], C> = DataPayloadCreator::default();
    |                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::default::Default` is not implemented for `[u8; 255]`
```

plaes:
Firstly, `GenericArray` dependency is used both in our code and our dependencies, whenever `default-crypto` is used.

### Our unconditional usage
* `JoinAcceptCreator.encrypt_payload(...)`: https://github.com/lora-rs/lora-rs/blob/969ee2406d957954921d4be6fda69adac976534b/lorawan-encoding/src/creator.rs#L175-L179
* We use it in `securityhelpers::encrypt_frm_data_payload(...)`:
https://github.com/lora-rs/lora-rs/blob/969ee2406d957954921d4be6fda69adac976534b/lorawan-encoding/src/securityhelpers.rs#L61

* We re-export `generic_array` from securityhelpers: https://github.com/lora-rs/lora-rs/
blob/969ee2406d957954921d4be6fda69adac976534b/lorawan-encoding/src/securityhelpers.rs#L2

### Internal re-export usage
* `EncryptedJoinAcceptPayload.decrypt(...)`: https://github.com/lora-rs/lora-rs/blob/969ee2406d957954921d4be6fda69adac976534b/lorawan-encoding/src/parser.rs#L378-L379

* `Encrypter`, `Decrypter` and `Keys` traits: https://github.com/lora-rs/lora-rs/blob/969ee2406d957954921d4be6fda69adac976534b/lorawan-encoding/src/keys.rs#L187-L201

## Use by dependencies
Regarding dependencies, RustCrypto (from where our aes and cmac implementations are coming from) is trying to get rid of GenericArray dependency as well in favor of const generics - https://github.com/RustCrypto/traits/issues/1481

As an interim step they have introduced a [`hybrid_array` crate](https://docs.rs/hybrid-array/latest/hybrid_array) and have ported latest release candidate version to use it instead of GenericArray.

plaes:
I have a quick and dirty (with only minimal changes to get it to build) branch here that uses the pre-release crates: https://github.com/plaes/lora-rs/tree/hybrid-array

[...]

PR #290, opened 2024-09-20
Drop GenericArray requirement from lorawan-device

plaes:
This is first step towards #126, though I have some doubts whether this is implemented in acceptable way...

### DIFF
--- a/examples/nrf52840/build.rs
+++ b/examples/nrf52840/build.rs
@@ -82,7 +82,7 @@ fn main() {
     // Put linker configuration in our output directory and ensure it's
     // on the linker search path.
     if cfg!(feature = "link-to-ram") {
-        File::create(out.join("link_ram_cortex_m.x"))
+        File::create(out.join("link_ram.x"))
             .unwrap()
             .write_all(include_bytes!("../link_ram_cortex_m.x"))
             .unwrap();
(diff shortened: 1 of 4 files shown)


lucasgranberg:
Maybe this is a step in a grander plan but I don't see this making that big of a difference. Now lorawan-device requires DataPayloadArray which is a GenericArray anyway. Not having to have the right version of GenericArray is a plus but the same could be achieved with an re-export.

I am playing around with just ripping out generic-array from everywhere except default-crypto and using [u8;256] and [u8;16] instead. Only snaffoo is I can't seem to convert a GenericArray to an [u8].

https://github.com/lora-rs/lora-rs/compare/main...lucasgranberg:lora-rs:rip-generic-array

plaes:
@lucasgranberg Grand plan is this - https://github.com/lora-rs/lora-rs/issues/126#issuecomment-2357778569

lucasgranberg:
Yes but why even bother with GenericArray/HybridArray. We don't need generic lengths on our arrays as we know what they should be. Only reason GenericArray was used was for passing in to the crypto stuff. That can still be done inside default-crypto. We can just use Rust arrays and slices in the rest of the code.

plaes:
Handled in #309

PR #292, opened 2024-10-01
Add start_rx, wait_for_packet_start, and complete_rx

CBJamo:
This allows for more precise control of rx.

For example you can now start receiving, wait for the preamble irq, then if you rx the preamble, prevent another future from switching the radio to tx mode.

```rust
radio
    .prepare_for_rx(RxMode::Continuous, &mod_params, &packet_params)
    .await?;

radio.start_rx().await?;

debug!("waiting tx/rx");
match select(radio.wait_for_packet_start(), tx_recv.receive()).await {
    Either::First(res) => {
        debug!("rx");
        res?;

        let mut tmp = RadioBuffer::new();
        let _ = tmp.resize_default(MAX_PACKET_SIZE);

        match with_timeout(
            Duration::from_millis(PACKET_TIME_MILLIS * 2),
            radio.complete_rx(&packet_params, &mut tmp),
        )
        .await
        {
            Ok(res) => {
                let (len, _status) = res?;
                tmp.truncate(len as usize);

                let buff = rx_send.send().await;
                *buff = tmp;
                rx_send.send_done();
                debug!("rx done");
            }
            Err(_) => {
                debug!("Preamble detected, but no packet rxed");
            }
        }
    }
    Either::Second(buff) => {
        // TX buff, can have backoff, cad, etc here.
```


lora-phy/src/lib.rs
```diff
@@ -268,12 +300,23 @@ where
                         return Err(err);
                     }
                 }
+                self.wait_for_irq().await?;
```
lucasgranberg, line 303: What are the effects of moving this to the end of the loop? I guess we would match Ok(None) but I am not sure

lucasgranberg:
I like the change you made here. I think it could be used for lorawan class B also. I don't have the time to test on real hw right now. We now have `prepare_for_rx`, `rx`, `start_rx` and `complete_rx`. For somebody using this for the first time it might be a bit confusing but the examples are clear so it should be fine.

lora-phy/src/lib.rs
CBJamo: Matching Ok(None) is what happens, yeah.

I wish I'd been smart enough to comment that change, as I wrote this months ago and can't remember the motivation. I think it had something to do with the irq already being cleared if you'd called wait_for_packet_start, but I'll have to dig in and do some testing to be sure.

lucasgranberg: Generally I think it is probably better after, as you wrote it, to avoid edge cases. "Clearing the queue" so to say but this is just a gut feeling.

plaes (approved):
Works with sx127x driver and LoRaWAN.

plaes (approved):

PR #296, opened 2024-11-20
lora-phy: Add API for specifying the sync word

plaes:
This is manually rebased version of #288.


lucasgranberg:
Thanks @jeholliday and @plaes for making this happen :+1:

Issue #291, opened 2024-09-23
Make defmt optional?


plaes:
We currently use defmt quite liberally:
* [ ] lora-phy - required dependency (and therefore also requires `defmt` for lorawan-device - https://github.com/lora-rs/lora-rs/blob/5664d38c692da8574adc1dd353de29d307540ab4/lora-phy/Cargo.toml#L17-L21
* [ ] lorawan-encoding - enabled by default: https://github.com/lora-rs/lora-rs/blob/5664d38c692da8574adc1dd353de29d307540ab4/lorawan-encoding/Cargo.toml#L33-L39

Ideally it should be an opt-in dependency, instead of opt-out.

PR #297, opened 2024-11-21
make defmt optional

lucasgranberg:
fixes https://github.com/lora-rs/lora-rs/issues/291


CBJamo:
I would recommend tagging the feature (ie "defmt-03") with the version, so that if defmt makes a breaking change that requires a new impl, that impl can be gated behind a different feature.

lorawan-device/Cargo.toml
```diff
@@ -27,21 +26,35 @@ defmt = { version = "0.3", optional = true }
...
 embassy-time = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
-tokio = { version = "1", features = ["rt", "macros", "time", "sync"]}
+tokio = { version = "1", features = ["rt", "macros", "time", "sync"] }
 rand = { version = "0", features = ["getrandom"] }
 lazy_static = "1"
 # Pull in lorawan/default-crypto which is required for tests
-lorawan = { path = "../lorawan-encoding", features = ["default-crypto"] }
+lorawan = { path = "../lorawan-encoding", default-features = false, features = [
```
plaes, line 41: Can you check whether running `cargo test --all-features` works from lora-phy after this change?

lora-phy/Cargo.toml
```diff
@@ -14,17 +14,17 @@ all-features = true
...
-], optional = true }
+lorawan-device = { path = "../lorawan-device", default-features = false, version = "0.12", optional = true }
 num-traits = { version = "0.2", default-features = false }
 embedded-hal = { version = "1" }
 embedded-hal-async = { version = "1" }
 document-features = "0.2.8"
 
 [features]
+
+defmt = ["dep:defmt", "lorawan-device/defmt", "lora-modulation/defmt"]
```
plaes, line 27: Please also add description for the feature (eventually will improve API docs once we pull in [document-features](https://crates.io/crates/document-features) like we do for lorawan-device):
`##  Enable defmt logging framework`

lorawan-device/src/lib.rs
```diff
@@ -38,7 +41,7 @@ pub use rand_core::RngCore;
 mod rng;
 pub use rng::Prng;
 
-mod log;
+// mod log;
```
plaes, line 44: hum?

plaes (commented):
My first approach would've been making gradual changes per library for easier review:
1. lora-phy
2. lorawan-device
3. ..feature flags for other libs

lucasgranberg:
> I would recommend tagging the feature (ie "defmt-03") with the version, so that if defmt makes a breaking change that requires a new impl, that impl can be gated behind a different feature.

I don't know about putting the version in the feature flag, We have been using "defmt" this far. Should we use "serde-1" also? Should I change all occurrences of the feature flag "defmt" to "defmt-03"?

plaes:
> > I would recommend tagging the feature (ie "defmt-03") with the version, so that if defmt makes a breaking change that requires a new impl, that impl can be gated behind a different feature.
> 
> I don't know about putting the version in the feature flag, We have been using "defmt" this far. Should we use "serde-1" also? Should I change all occurrences of the feature flag "defmt" to "defmt-03"?

Yes, `defmt-03` feature flag.

Though some packages even go as far as to [redefine imported package as `defmt-03`](https://github.com/rust-embedded/embedded-hal/blob/master/embedded-hal-bus/Cargo.toml).

lthiery (approved):
I don't understand this versioning but if @CBJamo and @plaes think it's right I'm not gonna argue.

plaes (approved):
Looks good from my side.

plaes (approved):

plaes:
There's still at least one issue when lorawan-device is built with defmt-03, but lora-phy isn't:
```
error[E0277]: the trait bound `lorawan_radio::Error: defmt::traits::Format` is not satisfied
  --> lora-rs/lora-phy/src/lorawan_radio.rs:92:21
   |
92 |     type PhyError = Error;
   |                     ^^^^^ the trait `defmt::traits::Format` is not implemented for `lorawan_radio::Error`
   |
   = help: the following other types implement trait `defmt::traits::Format`:
             &T
             &mut T
             ()
             (T0, T1)
             (T0, T1, T2)
             (T0, T1, T2, T3)
             (T0, T1, T2, T3, T4)
             (T0, T1, T2, T3, T4, T5)
           and 206 others
note: required by a bound in `lorawan_device::async_device::radio::PhyRxTx::PhyError`
  --> lora-rs/lorawan-device/src/async_device/radio.rs:34:20
   |
34 |     type PhyError: defmt::Format;
   |                    ^^^^^^^^^^^^^ required by this bound in `PhyRxTx::PhyError`
```

lucasgranberg:
> There's still at least one issue when lorawan-device is built with defmt-03, but lora-phy isn't

Good catch! I removed the defmt requirement as discussed in chat

plaes (approved):
Seems to work with my custom code now that defmt requirement from PhyRxTx trait was dropped.

lulf (approved):

PR #299, opened 2024-11-29
lorawan-encoding: Rename NewSKey to NwkSKey

plaes:
Should be mostly backwards compatible.

### DIFF
--- a/lorawan-device/CHANGELOG.md
+++ b/lorawan-device/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) 
 
 ## Unreleased
 
+- Deprecate NewSKey in favor of more commonly used NwkSKey
 - Rename the defmt feature to defmt-03
 
 ## [v0.12.1]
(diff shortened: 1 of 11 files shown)


lthiery (approved):
Fine by me pending clippy/fmt

plaes:
First step is done, lorawan-device will be somewhat more work as `newskey` is also public member of some structs..

plaes:
And second step is also done. But `newskey` is still used as public member of struct in at least 3 places:

https://github.com/lora-rs/lora-rs/blob/97addb9b385a01c9e64acb2d2cb1222ddf92812a/lorawan-device/src/mac/session.rs#L36-L40

https://github.com/lora-rs/lora-rs/blob/97addb9b385a01c9e64acb2d2cb1222ddf92812a/lorawan-device/src/mac/session.rs#L24-L32

https://github.com/lora-rs/lora-rs/blob/97addb9b385a01c9e64acb2d2cb1222ddf92812a/lorawan-device/src/lib.rs#L75-L79

lthiery:
I doubt many folks are using these newskey interfaces. I'd be OK with a breaking change and a version bump next time we do a release.

lthiery (approved):

PR #304, opened 2024-12-06
lorawan-encoding: Drop `is_uplink` information from mac commands

plaes:
Bunch of smaller updates:
* Split out maccommand macros into separate mod
* Drop uplink information, now that commands are split separately this info is not needed anymore
* Add back MSB/LSB info to docstrings
* Drop some disabled code
* ...and fix `mac_cmds_enum` macro to allow correct enum syntax


lucasgranberg:
msb/lsb doesn't make sens in this context I think? I guess we need to specify endianness if  anything

lthiery (approved):
Leave some for the rest of us! <3

lthiery:
> msb/lsb doesn't make sens in this context I think? I guess we need to specify endianness if anything
I think endianness and MSB/LSB statements are equivalent? all I know is that iin my experience, t's been super frustrating for folks making sure the keys from the LNS and the device match so I want to make sure we are clear about the specification. Fortunately, I think all hex encodings are MSB/big-endian. It's the byte arrays that flip-flop around all the time

Issue #303, opened 2024-12-06
LoRaWAN certification protocol  (TS009-1.2.1) encoding/decoding


plaes:
In order to certify LoRaWAN 1.0.4 (and 1.1) devices, firmware needs to pass the certification protocol, which is implemented as LoRaWAN communication over `fport = 224`.

Also, having these commands implemented, allows us to introduce our own "homegrown" HIL testing framework.

Required commands:
* [ ] 0x00 - PackageVersionReq / PackageVersionAns - Package version implemented by the end-device
* [ ] 0x01 - DutResetReq  - Reset the Microcontroller Unit
* [ ] 0x02 - DutJoinReq -  Reset the LoRaWAN MAC layer and start issuing Join-Request messages
* [ ] 0x03 - SwitchClassReq - Change class of operation to A, B, or C
* [ ] 0x04 - AdrBitChangeReq - Activate/deactivate Adaptive Data Rate (ADR)
* [ ] 0x05 - RegionalDutyCycleCtrlReq - Activate/deactivate the regional band duty-cycle enforcement
* [ ] 0x06 - TxPeriodicityChangeReq - Change uplink periodicity to the provided value
* [ ] 0x07 - TxFramesCtrlReq - Send all subsequent uplinks of the specified type
* [ ] 0x08 - EchoPayloadReq / EchoPayloadAns - echo the provided payload, where each byte is incremented by 1
* [ ] 0x09 - RxAppCntReq / RxAppCntAns - Provide the current applicative RxAppCnt value
* [ ] 0x0A - RxAppCntResetReq - Reset the applicative RxAppCnt value to 0
* 0x0B-0x1F RFU
* [ ] 0x20 - LinkCheckReq - Send a LinkCheckReq MAC command
* [ ] 0x21 - DeviceTimeReq - Send a DeviceTimeReq MAC command
* [ ] 0x22 - PingSlotInfoReq - Send a PingSlotInfoReq MAC command (only required for Class B DUTs)
* 0x23-0x3F RFU
* [ ] 0x40 - BeaconRxStatusIndCtrl - Activate/deactivate the autonomous BeaconRxStatusInd transmission
* [ ] 0x41 - BeaconRxStatusInd -  Autonomously send indication every time a beacon is supposed to have been received
* [ ] 0x42 - BeaconCntReq - Request to provide the current BeaconCnt value
* [ ] 0x43 - BeaconCntAns -  Return answer to the BeaconCntReq request
[...]

PR #305, opened 2024-12-09
Use proc_macros to generate lorawan-encoding boilerplate

plaes:
When I started looking into implementing the LoRaWAN testing protocol commandset ( #303 ), I didn't like reusing existing macros and instead decided to commit crimes of my own..

Long term idea for defining Mac commands and payload structures in single location, and this will create both commands/payloads and builders as well:

```rust
#[derive(CommandHandler)]
enum DownlinkCommand {
  #[cmd(cid=0x12, len=2)]
  CommandReq(CommandReqPayLoad<'a>)
}
```

And this would generate all the common boilerplate:
```rust
impl DownlinkCommand { ... }
impl DowlinkCommandCreator{ ... }

impl CommandReq { ... }
impl CommandReqCreator { ... }
impl CommandReqPayload { ... }
impl SerializableMacCommand for CommandReqCreator { ... }
```

Todo list:

* [x] Get rid of `mac_cmds_enum` macro in favor of `#[derive(CommandHandler)]` macro
* [x] Replace `mac_cmd_zero_len` macro to use attributes annotations: `#[cmd(cid=0x06, len=0)]` (for zero-length payloads)
* [x] Docstring handling
* [x] Replace `mac_cmds` to attribute-based derivation - non-zero length payloads 
* [x] `maccommandcreator` support... 
* [x] Generate `UplinkMacCommandCreator`/`DownlinkMacCommandCreator` from `CommandHandler` macro

For later:
* [ ] Migrate `create_ack_fn` and `create_value_reader_fn` to attribute-based macros?
* [ ] Investigate `maccommands::Error` usage (use `BufferTooShort(cid)` in macros) and get rid of unused variants) 
* [ ] Get rid of `&[&dyn SerializableMacCommand]` argument
* [ ] Split out MacCommandIterator to separate `derive(...)` ?


lthiery:
I haven't had time to look at what you've done here yet but I'm excited to take a look! It sounds like a promising approach

plaes:
Most of the heavy lifting is now done, so I would rather handle the last two items in the current TODO list as separate PRs.

plaes:
@lthiery @lucasgranberg How does it feel?

Also worth noting that net code change is actually negative (even with additional documentation):
` 8 files changed, 604 insertions(+), 631 deletions(-)`

lucasgranberg:
I like it, looks cleaner at least on the surface. Better documented too. Hard for me to say anything about the proc_macro itself as I have not written that many myself.

lorawan-macros/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+                    #[derive(Debug, PartialEq, Eq)]
+                    #[cfg_attr(feature = "defmt-03", derive(defmt::Format))]
+                    #( #doc )*
+                    pub struct #t <#lt> (pub(crate) &#lt [u8]);
+
+                    impl<#lt> #t<#lt> {
+                        /// Creates a new instance of the MAC command if there is enought data.
+                        pub fn new(data: &#lt [u8]) -> Result<#t<#lt>, Error> {
+                            if data.len() != #len {
+                                Err(Error::BufferTooShort)
```
lthiery, line 129: could we provide self.cid() here to help with possible debug?

lorawan-macros/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+            // Payloads with len > 0 (which have lifetime)
+            if let Some(lt) = lt {
+                payload_struct_impls.push(quote! {
+                    #[derive(Debug, PartialEq, Eq)]
+                    #[cfg_attr(feature = "defmt-03", derive(defmt::Format))]
+                    #( #doc )*
+                    pub struct #t <#lt> (pub(crate) &#lt [u8]);
+
+                    impl<#lt> #t<#lt> {
+                        /// Creates a new instance of the MAC command if there is enought data.
```
lthiery, line 126:
```suggestion
                        /// Creates a new instance of the MAC command if there is enough data.
```

lorawan-macros/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+                    fn payload_bytes(&self) -> &[u8] {
+                        &self.build()[1..]
+                    }
+
+                    /// The cid of the SerializableMacCommand.
+                    fn cid(&self) -> u8 {
+                        self.build()[0]
+                    }
+
+                    /// Length of the SerializableMacCommand without the cid.
```
lthiery, line 219:
```suggestion
                    /// Length of the SerializableMacCommand not including CID.
```

lorawan-macros/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+                        #len + 1
+                    }
+                }
+
+                impl SerializableMacCommand for #payload_creator {
+                    fn payload_bytes(&self) -> &[u8] {
+                        &self.build()[1..]
+                    }
+
+                    /// The cid of the SerializableMacCommand.
```
lthiery, line 214:
```suggestion
                    /// The CID of the SerializableMacCommand.
```

lthiery (approved):
I left a few nits about the comments along the way (primarily, probably good to homogenize on CID being capitalized in comments (I know you didn't inherit that)) but this is great so far. But isn't this very close to meaning that
```
pub fn can_piggyback(cmds: &[&dyn SerializableMacCommand]) -> bool``` 
```
can become
```
pub fn can_piggyback(cmds: &[SerializableMacCommand]) -> bool
```
Any big challenges in doing that? We could then make similar changes in lorawan-device. Perhaps you were trying to maintain the old API which makes sense!

lthiery:
FWIW, I had implemented the multicast messages [with the old macro system](https://github.com/lora-rs/lora-rs/commit/0a2bffcc9b2967c8e633a2daa3bf59721be47943). I think this will work well with what you've done here.

lorawan-macros/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "2.0"
+quote = "1.0"
+proc-macro2 = "1"
+
+# For "println" debugging via eprintln!("{:?}", ...);
+#syn = { version = "2.0", features = ["extra-traits"] }
```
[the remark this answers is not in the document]
plaes: This is only useful when debugging code generation via `cargo expand maccommands`. Therefore I would keep it as is.

lorawan-macros/src/lib.rs
plaes: Yes.. and that error enum contains unused stuff as well.. ;)

lthiery (approved):

PR #306, opened 2024-12-09
Move common internal data types (frequency, dlsettings, etc) from maccomands to `lorawan::types`

plaes:
This PR moves most internal types used inside payloads from `maccommands` module to common `types` module.
No functional changes, and moves are made in a backwards compatible way to preserve API. (Unfortunately deprecated attribute doesn't really work when re-exporting).
This is also a preparation PR for #305


lthiery (approved):

lthiery:
Nice work

PR #307, opened 2024-12-16
lora-phy: fix Sx126x::process_irq_event

fe1es:
The `LoRa::cad` always returns true because the `Sx126x::process_irq_event` doesn't check the IrqMask::CADActivityDetected flag when the IrqMask::CADDone flag is set. This fixes the issue.


plaes (approved):
Oof, nice catch and apologies as I introduced this in 7dcb66014bff1f6fbb89f9b04eb9f4a0cf5f109c.

PR #309, opened 2024-12-17
Remove generic array

lucasgranberg:
With this change generic-array is removed as an requirement for lorawan-device and for anyone not using default-crypto. This is an alternative to https://github.com/lora-rs/lora-rs/pull/290 and would fix
https://github.com/lora-rs/lora-rs/issues/126


lorawan-encoding/src/securityhelpers.rs
```diff
@@ -58,17 +56,16 @@ pub fn encrypt_frm_data_payload(
...
     let mut s = [0u8; 16];
-    let s_block = GenericArray::from_mut_slice(&mut s[..]);
 
     let mut ctr = 1;
     for i in 0..len {
         let j = i & 0x0f;
         if j == 0 {
             a[15] = ctr;
             ctr += 1;
-            s_block.copy_from_slice(&a);
```
lthiery, line 69: `s = a` means we can just use `a` and skip `s` entirely?

lthiery (approved):
This looks well done to me! Clean work

plaes (commented):
My main concern currently is usage of `[u8; 256]` as common type for both lorawan-device and lorawan-encoding.
Just wondering wheter we should use type alias for that in a similar way I did in#290 by defining the `DataPayloadArray`.

lucasgranberg:
> My main concern currently is usage of `[u8; 256]` as common type for both lorawan-device and lorawan-encoding.
> Just wondering wheter we should use type alias for that in a similar way I did in#290 by defining the `DataPayloadArray`.

I could do that. Taking it further, do we even need the length to be generic?

plaes:
> > My main concern currently is usage of `[u8; 256]` as common type for both lorawan-device and lorawan-encoding.
> > Just wondering wheter we should use type alias for that in a similar way I did in#290 by defining the `DataPayloadArray`.
> 
> I could do that. Taking it further, do we even need the length to be generic?

I don't think so. We eventually need to "cut" it if it exceeds maximum allowed length though.

lorawan-encoding/src/securityhelpers.rs
lucasgranberg: The problem would be that we would re-encrypt the "helper_block" ie `a` on long messages. I tried it and it makes the "test_complete_long_dataup_payload_frm_payload" test fail which makes sense.

lucasgranberg:
Actually looking at it now I only use [u8; 256] in the default-crypto `new` impl and that one is only used for tests. The creators uses an AsMut generic which is fine I think.

DataPayloadCreator however checks if the buffer is shorter than 255 which makes anything other than [u8;255] pretty useless. I guess the check should be for whatever the shortest message would be instead. Also adding checks when building the message with a payload.

plaes (approved):

PR #315, opened 2025-01-03
add length checks for creator buffers

lucasgranberg:
When using the payload creator the length of the target buffer was not checked that well.

### DIFF
--- a/lorawan-encoding/src/creator.rs
+++ b/lorawan-encoding/src/creator.rs
@@ -7,6 +7,10 @@ use super::maccommandcreator;
 use super::maccommands::{mac_commands_len, SerializableMacCommand};
 use super::parser;
 use super::securityhelpers;
+use crate::packet_length::phy::join::{
+    JOIN_ACCEPT_LEN, JOIN_ACCEPT_WITH_CFLIST_LEN, JOIN_REQUEST_LEN,
+};
+use crate::packet_length::phy::{MIC_LEN, PHY_PAYLOAD_MIN_LEN};
 use crate::types::{DLSettings, Frequency};
 
 #[cfg(feature = "default-crypto")]
@@ -39,8 +43,10 @@ impl<D: AsMut<[u8]>, F: CryptoFactory + Default> JoinAcceptCreator<D, F> {
     ///
     /// TODO: Add more details & and example
     pub fn with_options(mut data: D, factory: F) -> Result<Self, Error> {
-        // length verification will occur during building
         let d = data.as_mut();
+        if d.len() < JOIN_ACCEPT_LEN {
+            return Err(Error::BufferTooShort);
+        }
         d[0] = 0x20;
         Ok(Self { data, with_c_f_list: false, encrypted: false, factory })
     }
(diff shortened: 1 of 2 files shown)


lorawan-encoding/src/creator.rs
```diff
@@ -39,8 +40,11 @@ impl<D: AsMut<[u8]>, F: CryptoFactory + Default> JoinAcceptCreator<D, F> {
     ///
     /// TODO: Add more details & and example
     pub fn with_options(mut data: D, factory: F) -> Result<Self, Error> {
-        // length verification will occur during building
         let d = data.as_mut();
+        // JoinNonce(3) NetID(3) DevAddr(4) DLSettings(1) RXDelay(1)
+        if d.len() < 12 {
```
plaes, line 45: `JOIN_ACCEPT_PAYLOAD_LEN` from packet_lengths

lorawan-encoding/src/creator.rs
```diff
@@ -23,6 +23,7 @@ pub enum Error {
 }
 
 const PIGGYBACK_MAC_COMMANDS_MAX_LEN: usize = 15;
+const MIC_LEN: usize = 4;
```
plaes, line 26:
```
use crate::packet_length::phy::MIC as MIC_LEN;
```

lorawan-encoding/src/creator.rs
```diff
@@ -129,6 +133,9 @@ impl<D: AsMut<[u8]>, F: CryptoFactory + Default> JoinAcceptCreator<D, F> {
             return Err(Error::InvalidChannelList);
         }
         let d = self.data.as_mut();
+        if d.len() < 15 + ch_list.len() * 3 {
```
plaes, line 136: `CF_LIST_LEN`? (Though in our case it's defined as 16).

lorawan-encoding/src/creator.rs
```diff
@@ -223,6 +230,7 @@ impl<D: AsMut<[u8]>, F: CryptoFactory> JoinRequestCreator<D, F> {
     /// Creates a well initialized JoinRequestCreator with specific crypto functions.
     pub fn with_options(mut data: D, factory: F) -> Result<Self, Error> {
         let d = data.as_mut();
+        //JoinEUI(8) DevEUI(8) DevNonce(2)
         if d.len() < 23 {
```
plaes, line 234: `JOIN_REQUEST_PAYLOAD_LEN` 

Also missing space in comment

lorawan-encoding/src/creator.rs
```diff
@@ -320,7 +328,8 @@ impl<D: AsMut<[u8]>, F: CryptoFactory + Default> DataPayloadCreator<D, F> {
     /// By default the packet is unconfirmed data up packet.
     pub fn with_options(mut data: D, factory: F) -> Result<Self, Error> {
         let d = data.as_mut();
-        if d.len() < 255 {
+        //MHDR(1) DevAddr(4) FCtrl(1) FCnt(2) MIC(4)
+        if d.len() < 12 {
```
plaes, line 332: `PHY_PAYLOAD_MIN_LEN`?

Also, comment formatting (missing space).

plaes (changes requested):
Does this fix existing bug?

And how hard would be to construct a failing testcase (before the fixes)?

lorawan-encoding/src/creator.rs
lucasgranberg: you are right, it is always 16 bytes (both type 0 and 1). It is possible to pass in something shorter due to `AsRef<[Frequency<'a>]>` but that is another issue

lucasgranberg:
I forgot about the constants already created. Switched to using them now and removed the comments as the consts document themselves. tnx plaes

lucasgranberg:
> Does this fix existing bug?
> 
> And how hard would be to construct a failing testcase (before the fixes)?

It fixes some annoyances I had when using lorawan-encoding directly. Also there was `// length verification will occur during building` which is not true. It would panic with a zero length buffer. I don't think we need a test for that but I could add some to verify JoinAcceptCreator with cflist and DataPayloadCreator with maccommands + payload

lucasgranberg:
Already found some more mistakes I made. Also the set_c_f_list only works with 1.0.2 and doesn't set the last byte to 0 to indicate cflist type. Also build with cflist is broken.

These are not important for devices anyway but Ill fix them now while I am at it

plaes (approved):

lthiery (approved):

PR #317, opened 2025-01-03
Implement Multicast Messages

lthiery:
I've re-implemented the multicast messages from an old branch of mine using the attribute macro.

There is one tricky command that I forced me to make some changes to the attribute macro: `McGroupStatusAnsPayload`. It is a **variable length message**, therefore, I needed to make the fixed-length concept optional in the attribute macro.

To make this work right, I renamed `const fn len()` to become `const fn max_len()`, so that it could be used for payload creation in the case of variable length mac commands. `fn len(&self)` takes the old name, but also a reference to self so that it would be appropriate for variable length commands in the case the parsing.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,8 @@
-/Cargo.toml @ivajloip @lthiery @lulf
-/README.md @ivajloip @lthiery @lulf
+/Cargo.toml @ivajloip @lthiery @lulf @plaes
+/README.md @ivajloip @lthiery @lulf @plaes
 
 /lora-modulation/ @lthiery @lulf @ivajloip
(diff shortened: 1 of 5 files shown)


lthiery:
By the way, I consider this a WIP and intend on adding some test payloads once I can generate some from an LNS. I've also added @plaes as code owners to a few sections as I think he's gotten quite familiar with a few sections.

plaes (approved):
> There is one tricky command that I forced me to make some changes to the attribute macro: McGroupStatusAnsPayload. It is a variable length message, therefore, I needed to make the fixed-length concept optional in the attribute macro.

Thanks, I faced actually similar issue when implementing the certification protocol.

Issue #313, opened 2024-12-30
Make Class-C optional?


martin2250:
Hi, would you be interested in (aka accept a pull request) putting the LoRaWAN Class-C features behind a (default) feature flag?

I'm using lora-rs on a rather memory-constrained device (64K STM32) and the class-c code alone uses about ~10kB (according to cargo-bloat, specifically one closure in `between_windows`. By removing all class-C code (and panic-probe plus some others), I just managed to fit it in.

It would be nice to have a feature flag that could be used instead.

Also, in the same spirit, a transmit-only feature could be useful for ABP devices. I'm guessing, however, that this would require feature-gating a lot more sections of code.

plaes:
Yes, when disabling Class C drops ~10kB, adding a feature flag for it makes sense.

Also, not against the transmit-only feature, but our current priority is getting MAC+application layer properly work together to support things required for full 1.0.4 support (ADR, channel and timing management, etc see #302).

martin2250:
nvm, while working on the PR I found that the 10kB drop was not caused by class-c. I'm guessing what I saw was only panic-probe and defmt features in other crates.

The actual difference with / without class-c is only around 1kB and therefore probably not worth the complexity.

Cheers!

lthiery:
I was going to say, I was surprised that Class C would add such overhead since it is mostly a procedural change in listening and times when otherwise we'd been sleeping! I'd be open to feature-gating it though as I am working on a feature-gated multicast setup as well.

martin2250:
I still have the code in my fork (not 100% ready), I'll go over it later and open a PR if you'd still take it.

PR #318, opened 2025-01-04
lorawan-device: feature-gate class-c behavior

martin2250:
Hi,

as discussed in issue https://github.com/lora-rs/lora-rs/issues/313, here my implementation of the class-c feature flag.

Some code is written in an odd way (especially the start of `async fn between_windows(...)`. That was necessary because otherwise the compiler would not actually optimize away the dead code. I tried just using cfg!() in the `if`, but that didn't work.

The exact difference in the size of the `.text` section in my STM32L0 codebase is 1016 bytes.

My testing was done with
```toml
[profile.release]
opt-level = "z"
lto = true
debug = true
codegen-units = 1
```

Cheers!

### DIFF
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -315,7 +315,7 @@ where
         use self::radio::RxQuality;
         use futures::{future::select, future::Either, pin_mut};
 
-        if self.class_c {
+        if !self.class_c {
             self.radio.low_power().await.map_err(Error::Radio)?;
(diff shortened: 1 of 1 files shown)


lthiery:
I agree, `async fn between_windows(...)` did not look great. I opted for another approach and [opened a PR](https://github.com/martin2250/lora-rs/pull/1) against your branch. 

Otherwise, this looks good to me!

martin2250:
Looks good, this solution is more elegant than my mess or duplicating that code within a single function.

PR #320, opened 2025-01-07
Update to embassy-time-0.4

plaes:
Unfortunately we need to update everything in one go...

Looks like stm32 examples haven't been updated in a while and there are bunch of API changes in embassy-stm32

* [x] lorawan-device
* [x] examples/nrf52840
* [x] examples/rp
* [x] stm32l0
* [x] stm32wl

### DIFF
--- a/lorawan-device/Cargo.toml
+++ b/lorawan-device/Cargo.toml
@@ -30,7 +30,7 @@ serde = { version = "1", default-features = false, features = [
 ], optional = true }
 seq-macro = "0.3.5"
 document-features = "0.2.8"
-embassy-time = { version = "0.3.0", optional = true }
+embassy-time = { version = "0.4", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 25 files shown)


lorawan-device/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde = { version = "1", default-features = false, features = [
 ], optional = true }
 seq-macro = "0.3.5"
 document-features = "0.2.8"
-embassy-time = { version = "0.3.0", optional = true }
+embassy-time = { version = "0.4", optional = true }
```
lthiery, line 33:
```suggestion
embassy-time = { version = ">=0.3.0", optional = true }
```

it looks like  we could be more flexible here?

Dirbaio:
`embassy-stm32` v0.2.0 is now out, you could remove the `[patch.crates-io]`s :)

lthiery (approved):
looks like we are gearing up for a batch of breaking changes

PR #323, opened 2025-01-16
encoding: Misc documentation-wise formatting fixes and cleanups

plaes:
* Add `#[allow(clippy::len_without_is_empty)]` next to the len functions and drop global exception
* Clean up formatting inside docstrings
* Drop unneeded `clippy::upper_case_acronyms` exception


lorawan-encoding/src/maccommandcreator.rs
```diff
@@ -144,11 +144,8 @@ impl LinkADRReqCreator {
 ///
 /// ```
 /// let mut creator = lorawan::maccommandcreator::LinkADRAnsCreator::new();
-/// let res = creator
-///     .set_channel_mask_ack(true)
-///     .set_data_rate_ack(true)
-///     .set_tx_power_ack(true)
-///     .build();
+/// let res =
+///     creator.set_channel_mask_ack(true).set_data_rate_ack(true).set_tx_power_ack(true).build();
```
lthiery, line 148: did these get formatted this way by cargo fmt or did you prefer it this way? i personally prefer the way it was before but i don't care much either way

lthiery (approved):

lorawan-encoding/src/maccommandcreator.rs
plaes: I did it automatically using:
```
cargo +nightly fmt -- --unstable-features --config format_code_in_doc_comments=true
```

lthiery: Cool - let's gooo then

PR #324, opened 2025-01-17
500kHz Sensitivity optimization

powturns:
Support optimizations for 500kHz bandwidth as described in Errata 2.1.

The Errata indicates that this only applies to chips that have 0x12 in RegVersion, but I believe that is the case for all production chips.

See: https://github.com/sparkfun/SparkFun_LoRaSerial/blob/main/Documents/SX1276_77_8_ErrataNote_1.1_STD.pdf

### DIFF
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -58,6 +58,7 @@ pub struct Config<C: Sx127xVariant> {
 pub struct Sx127x<SPI, IV, C: Sx127xVariant + Sized> {
     intf: SpiInterface<SPI, IV>,
     config: Config<C>,
+    data: C::Data,
 }
 
 impl<SPI, IV, C> Sx127x<SPI, IV, C>
@@ -69,7 +70,11 @@ where
     /// Create an instance of the RadioKind implementation for the LoRa chip kind and board type
     pub fn new(spi: SPI, iv: IV, config: Config<C>) -> Self {
         let intf = SpiInterface::new(spi, iv);
-        Self { intf, config }
+        Self {
+            intf,
+            config,
+            data: Default::default(),
+        }
     }
 
     // Utility functions
@@ -125,6 +130,9 @@ where
         self.write_register(Register::RegSyncWord, sync_word).await?;
 
         self.set_tx_rx_buffer_base_address(0, 0).await?;
+
+        C::init_lora(self, sync_word).await?;
+
         Ok(())
     }
 
--- a/lora-phy/src/sx127x/sx1276.rs
+++ b/lora-phy/src/sx127x/sx1276.rs
@@ -12,7 +12,28 @@ use lora_modulation::Bandwidth;
 /// Sx1276 implements the Sx127xVariant trait
 pub struct Sx1276;
 
+#[derive(Default)]
+pub struct Sx1276Data {
+    /// flag to indicate the Errata 2.1: Sensitivity optimization with 500 kHz bandwidth is required
+    sensitivity_quirk: bool,
+}
+
 impl Sx127xVariant for Sx1276 {
+    type Data = Sx1276Data;
+
+    async fn init_lora<SPI: SpiDevice<u8>, IV: InterfaceVariant>(
+        radio: &mut Sx127x<SPI, IV, Self>,
+        _sync_word: u8,
+    ) -> Result<(), RadioError> {
+        let chip_version = radio.read_register(Register::RegVersion).await?;
+        debug!("Detected sx1276 v{}", chip_version);
+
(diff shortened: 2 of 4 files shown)


lthiery:
Is it not possible to check reg version before applying the optimization? or perhaps reg version may be read during setup and referred to during `set_modulation_params`

powturns:
I'll experiment with what it would look like to read it during setup

lora-phy/src/sx127x/mod.rs
```diff
@@ -58,6 +58,7 @@ pub struct Config<C: Sx127xVariant> {
 pub struct Sx127x<SPI, IV, C: Sx127xVariant + Sized> {
     intf: SpiInterface<SPI, IV>,
     config: Config<C>,
+    chip_version: Option<u8>,
```
plaes, line 61: Firstly, lets use `sensitivity_quirk: false` as this sensitivity errata seems to be only needed for `sx127{6,7,8}` and not sx1272.

So, we would need to go through chip-specific functions:

```rust
self.sensitivity_quirk = C::is_errata_needed().await?;
```

Where sx1272 function returns just `false` and sx1276 specific version looks something like this:
```rust
{
  let chip_version = self.read_register(Register::RegVersion).await?;
  debug!("Detected sx127x v{}", chip_version);
  chip_version == 0x12
};
```

lora-phy/src/sx127x/sx1276.rs
```diff
@@ -108,7 +108,32 @@ impl Sx127xVariant for Sx1276 {
         }
         let mut config_3 = radio.read_register(Register::RegModemConfig3).await?;
         config_3 = (config_3 & 0xf3u8) | ldro_agc_auto_flags;
-        radio.write_register(Register::RegModemConfig3, config_3).await
+        radio.write_register(Register::RegModemConfig3, config_3).await?;
+
```
plaes, line 112: We have basically 3 cases here (if quirk is needed):
``` rust
if is_quirk_needed {
  if mdltn_params.bandwidth == Bandwidth::_500KHz {
    let val_1 = 0x02;
    match mdltn_params.frequency {
       862..1020 => {
         write(val_1) to addr 0x36;
         write(0x64) to addr 0x3a;
         return Ok(())
      },
       410..525 => {
         write(val_1) to addr 0x36;
         write(0x7f) to addr 0x3a;
         return Ok(())
       }
      // fall through...
       _ = {}
  }  
  write(0x03) to addr 0x36
  // register 0x3a is handled automatically by chip...
}
``

**Review by plaes — CHANGES_REQUESTED** [review]

File: lora-phy/src/sx127x/mod.rs (hunk 1 above)
Comment (powturns) [reply lora-phy/src/sx127x/mod.rs]: took a slightly different approach to see what you think. I like the use of associated types because it doesn't leak the existence of 1276 implementation details to the 1272 chip, as well as provides a generic way for variants to store additional data if required. However, it does make the types a bit more complicated, and relies on the data type implementing default.

File: lora-phy/src/sx127x/sx1276.rs [hunk 3]
```diff
@@ -108,7 +129,48 @@ impl Sx127xVariant for Sx1276 {
...
+        if radio.data.sensitivity_quirk {
+            // apply Errata 2.1: Sensitivity optimization with 500 kHz bandwidth
+
+            if mdltn_params.bandwidth == Bandwidth::_500KHz {
+                let val_1 = 0x02;
+                match mdltn_params.frequency_in_hz {
+                    862..=1020 => {
```
powturns, line 140: Its unclear to me if the ranges in the errata are inclusive. Thoughts?

plaes: Firstly, please fix the frequency, currently these ranges are presented in kilohertz, but we should be comparing against hertz (just add extra 3 zeroes).

And inclusive is ok, though datasheets don't really say whether it is or isn't...

lora-phy/src/sx127x/sx1276.rs
```diff
@@ -108,7 +129,48 @@ impl Sx127xVariant for Sx1276 {
...
+                            .write_register(Register::RegHighBwOptimize2, 0x7f)
+                            .await?;
+                        return Ok(())
+                    }
+                    // fall through...
+                    _ => {}
+                }
+            }
+
+            // for all other combinations of bandwidth / frequencies, reset to 0x03 and the value at
```
plaes, line 164: Please clean up the comment.

And `cargo fmt` should clean up the extra newlines.

plaes (changes requested):
Please fix the frequencies.

plaes (approved):
Thanks!

PR #325, opened 2025-01-17
Unconditionally clear IRQ flags after config

powturns:
After writing the irq parameters, clear any currently active flags. According to the spec, it doesn't seem valid to ever write 0x00 to RegIrqFlags as this appears to be a no-op.


lthiery:
The writing of 0x00 does seem strange. This LGTM but I don't know the radio code very well.

plaes:
Indeed, these register fields are `rc` where `r` is Read and `c` is "**set to clear**". And each bit even has description "writing a 1 clears the IRQ" ;)

I can see no functional changes on sx1276 radio (at least in P2P mode), but I'll need to also test with sx1272 and LoRaWAN.

plaes (approved):
Thanks

Tested on sx1272 and sx1276 radios.

PR #326, opened 2025-01-24
lora-phy: Errata cleanup

plaes:
After merging #324 I noticed that we can actually utilize the chip struct itself for keeping the chip-specific data.

cc: @powturns


powturns:
I thought about doing it this way but thought it was less clean having to create a chip when passing it into the config. I don't have a strong preference either way though.

plaes:
> I thought about doing it this way but thought it was less clean having to create a chip when passing it into the config. I don't have a strong preference either way though, and this approach does have the advantage of allowing the user to specify chip-specific parameters (I dont know if there is a use case for that)

Yeah, unfortunately it looked good initially, but then turns out I would need to initialize sx1276 as object..

PR #328, opened 2025-01-30
Fix DioMapping1Dio1 Shifts

powturns:
According to the datasheet, Dio1 occupies bits 5-4 (zero indexed) of the register. This requires a left shift of 4 to skip over Dio3 and Dio2

image


plaes (approved):
Aha, nice catch.

PR #329, opened 2025-01-30
Configure RxTimeout interrupt on Dio1

powturns:


plaes (approved):
Tested with sx1272 radio.

PR #332, opened 2025-02-03
device: Move downlink mac handler from Configuration to mac

plaes:
Downlink handling should be done at least at mac scope.

[Edit]: Meh, it's not really at Mac scope, but in Session scope, though it's an improvement.


lthiery:
> [Edit]: Meh, it's not really at Mac scope, but in Session scope, though it's an improvement.

Yea, I think that's correct though. MAC commands don't make sense without a session.

lthiery (approved):

PR #334, opened 2025-02-06
Certification protocol implementation

plaes:
This adds partial list of payloads of Certification protocol to lorawan-encoding.
These together with bunch of glue code added to our lorawan-device's mac + async_device layer is good enough to pass the Activation Pre-Test for class A and C devices (at least when testing with virtual-device).

### DIFF
--- a/lorawan-encoding/src/certification.rs
+++ b/lorawan-encoding/src/certification.rs
@@ -17,6 +17,10 @@ pub enum DownlinkDUTCommand<'a> {
     #[cmd(cid = 0x02, len = 0)]
     DutJoinReq(DutJoinReqPayload),
 
+    /// Request to activate/deactivate Adaptive Data Rate (ADR)
+    #[cmd(cid = 0x04, len = 1)]
+    AdrBitChangeReq(AdrBitChangeReqPayload<'a>),
+
     /// Change uplink periodicity to the provided value
     #[cmd(cid = 0x06, len = 1)]
     TxPeriodicityChangeReq(TxPeriodicityChangeReqPayload<'a>),
@@ -31,6 +35,17 @@ pub fn parse_downlink_certification_messages(
     data: &[u8],
 ) -> MacCommandIterator<'_, DownlinkDUTCommand<'_>> {
     MacCommandIterator::new(data)
+}
+
+impl AdrBitChangeReqPayload<'_> {
+    /// Enable/disable ADR
+    pub fn adr_enable(&self) -> Result<bool, Error> {
+        match self.0[0] {
+            0 => Ok(false),
+            1 => Ok(true),
+            _ => Err(Error::RFU),
+        }
+    }
 }
 
 impl TxPeriodicityChangeReqPayload<'_> {
--- a/lorawan-encoding/src/certification.rs
+++ b/lorawan-encoding/src/certification.rs
@@ -1,6 +1,7 @@
 //! LoRaWAN Certification Protocol (TS009) command and payload handling
 use lorawan_macros::CommandHandler;
 
+use crate::creator::UnimplementedCreator;
 use crate::maccommands::MacCommandIterator;
 use crate::maccommands::SerializableMacCommand;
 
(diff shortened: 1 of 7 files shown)


lorawan-encoding/src/certification.rs
```diff
@@ -0,0 +1,136 @@
...
+    const fn min_len() -> usize {
+        // Minimum length of the of payload including CID
+        2
+    }
+
+    /// Actual length of the payload
+    #[allow(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
+        // Payload should have at least minimum length...
+        core::cmp::max(Self::min_len(), self.0.len())
```
lthiery, line 120: I was concerned you made a mistake here but it looks like these commands are sent only one at a time! Interesting

lorawan-encoding/src/certification.rs
```diff
@@ -0,0 +1,136 @@
...
+pub enum UplinkDUTCommand<'a> {
+    #[cmd(cid = 0x7f, len = 12)]
+    DutVersionsAns(DutVersionsAnsPayload<'a>),
+}
+
+pub fn parse_downlink_certification_messages(
+    data: &[u8],
+) -> MacCommandIterator<'_, DownlinkDUTCommand<'_>> {
+    MacCommandIterator::new(data)
+}
```
lthiery, line 49: This API would lead to less confusion for the user
```suggestion
pub fn parse_downlink_certification_message(
    data: &[u8],
) -> Option<DownlinkDUTCommand<'_>> {
    let mut iter: MacCommandIterator<'_, DownlinkDUTCommand<'_>> = MacCommandIterator::new(data);
    iter.next()
}
```

lorawan-encoding/src/certification.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Actual length of the payload
+    #[allow(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
+        // Payload should have at least minimum length...
+        core::cmp::max(Self::min_len(), self.0.len())
+    }
+
+    /// Whether all subsequent uplinks are sent as L2 Unconfirmed (`FType = 2`),
+    /// L2 Confirmed (`FType = 4`) or switch back to device default.
+    pub fn frame_type(&self) -> Result<Option<bool>, Error> {
```
lthiery, line 125:
```suggestion
    pub fn frame_type_is_confirmed(&self) -> Result<Option<bool>, Error> {
```

lorawan-encoding/src/certification.rs
lthiery: Ideally, the attribute macro could give us a function to try to parse just the next message so we don't make this degenerate use of the iterator, but this was easy to do at least (and we can fix it later without changing the API)

lthiery (changes requested):
Looks go

lorawan-encoding/src/certification.rs
plaes: Yeah, this is annoying. Probably warrants a better comment and I'll also send a note about improving the protocol documentation.

lorawan-encoding/src/certification.rs
plaes: Ah yes, I initially returned `Option<FrameType>` instead of `Option<bool>`, but then ran into `serde` issues because this is eventually used in `mac::Session`.
Though, I guess better name would be actually `frame_type_override()` as this function returns actually 3 values:
1. do not override
2. force unconfirmed
3. force confirmed

lorawan-encoding/src/certification.rs
plaes: Let's leave this for future. We use this construct for all the commands and it's extensively used in tests as well.

plaes:
Pushed with fixes to review comment, except the `parse_downlink_certification_message` API change as this is something I would eventually do as a separate PR.

lthiery (approved):

PR #336, opened 2025-02-07
rewrite uplink mac commands

lthiery:
The previous `uplink::Uplink` strucure was not going to scale well. 

We maintain a small intermediate buffer of size FOPTS_MAX_LEN (we don't really support sending things on Port 0 yet).

The data payload creator has also been changing to just accept a slice (instead of  `&[&dyn SerializableMacCommand]`).


plaes (approved):

plaes (changes requested):
Looks great, though there are some issues still.

plaes (approved):

PR #343, opened 2025-02-12
provide function name in invalid radio mode error

lthiery:


lucasgranberg:
would it be possible to use the RadioMode enum instead of str?

lthiery:
I believe so but I'm using str so I can identify the exact fan that's erroring. I think if I use the enum, I still won't be able to disambiguate?

lora-phy/src/lib.rs
```diff
@@ -268,7 +268,7 @@ where
         if let RadioMode::Receive(listen_mode) = self.radio_mode {
             self.radio_kind.do_rx(listen_mode).await
         } else {
-            Err(RadioError::InvalidRadioMode)
+            Err(RadioError::InvalidRadioMode("start_tx"))
```
lucasgranberg, line 271: this should be "start_rx"

lucasgranberg:
Well you would see that the tx fn returns an error something like `InvalidRadioMode(RadioMode::Rx)`.

lucasgranberg:
Doing it with RadioMode could look like this too: 
```rs
pub enum RadioError {
    ...
    InvalidRadioMode { expected: RadioMode, actual: RadioMode },
    ...
}
Err(RadioError::InvalidRadioMode {
    expected: RadioMode::ChannelActivityDetection,
    actual: self.radio_mode,
})
```
Ofc this makes the RadioError 24 byte but idk if that is so bad. defmt should handle it quite well

I don't have any strong reasons for avoiding adding the fn name as a str to the Error other than it "feels wrong". I guess the error information tells us it errors bc we were in the wrong fn with the wrong mode and it should not be about the fn but about the modes. The difference is subtle.

PR #353, opened 2025-02-25
Drop datarates generic and use a table with 15 rates for all regions

plaes:
As datarate table has maximum 16 elements (and last one isn't really a datarate), carrying the generic const around seems a bit pointless...  And we'll be gaining another generic once we'll plug in RXParamsetupReq which featuring rx1 datarate offsets.

Changes to lorawan crate size using `cargo bloat`:
```
nrf52840_lorawan all_regions
before: 17.7KiB
after: 17.8KiB

nrf52840_lorawan only eu868
before: 8.6KiB
after:  8.6KiB

stm32wl_lorawan_class_c
before: 20.2KiB
after: 20.3KiB
```


lthiery (commented):
I think I went hog on generics in regions and this is just the beginning of fixing it. Well done 👏

lthiery (approved):

PR #354, opened 2025-03-02
RFC: RXParamSetup rx1_dr_offset support

plaes:
This is WIP PR (only implemented for eu868 region atm) because I need some help with the data design.

My initial approach was to add a new generic and then encode the DR offset table like this (note that we need all 15 rows):

```rust
impl ChannelRegion<NUM_DR_OFFSETS> for EU868Region {
    fn rx1dr_offsets() -> &'static [Option<[u8; NUM_DR_OFFSETS]>; NUM_DATARATES as usize] {
        &RX1DR_OFFSETS
    }
    // ...
}

pub(crate) const RX1DR_OFFSETS: [Option<[u8; NUM_DR_OFFSETS]>; NUM_DATARATES as usize] = [
    Some([0, 0, 0, 0, 0, 0]),
    Some([1, 0, 0, 0, 0, 0]),
    Some([2, 1, 0, 0, 0, 0]),
    Some([3, 2, 1, 0, 0, 0]),
    Some([4, 3, 2, 1, 0, 0]),
    Some([5, 4, 3, 2, 1, 0]),
    Some([6, 5, 4, 3, 2, 1]),
    None,
    None,
    None,
    None,
    None,
    None,
    None,
    None,
];
```

Second approach would be probably to add a function which does something like this:
```rust
fn get_offset(cur_dr, offset) {
match cur_dr {
   0 => [0, 0, 0, 0, 0, 0][offset],
   1 => [1, 0, 0, 0, 0, 0][offset],
   2 => [2, 1, 0, 0, 0, 0][offset],
   ...
   _ => None
}
```

### DIFF
--- a/lorawan-device/src/async_device/test/mod.rs
+++ b/lorawan-device/src/async_device/test/mod.rs
@@ -224,6 +224,7 @@ async fn test_confirmed_uplink_with_ack_rx2() {
 }
 
 #[tokio::test]
+#[ignore]
 async fn test_link_adr_ans() {
     let (radio, timer, mut async_device) = setup_with_session();
     let send_await_complete = Arc::new(Mutex::new(false));
--- a/lorawan-device/src/async_device/test/util.rs
+++ b/lorawan-device/src/async_device/test/util.rs
@@ -11,7 +11,8 @@ use crate::{AppSKey, NwkSKey};
 fn setup_internal(session_data: Option<Session>) -> (RadioChannel, TimerChannel, Device) {
     let (radio_channel, mock_radio) = TestRadio::new();
     let (timer_channel, mock_timer) = TestTimer::new();
-    let region = region::US915::default();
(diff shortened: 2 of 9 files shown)


lthiery:
Couldn't this be implemented with `(cur_dr).saturating_sub(offset)`? or a more generalized one would be `(cur_dr).saturating_sub(offset).min(min_dr)` (i think some plans have transmits DR that start at like 5 or something).

plaes:
> Couldn't this be implemented with `(cur_dr).saturating_sub(offset)`? or a more generalized one would be `(cur_dr).saturating_sub(offset).min(min_dr)` (i think some plans have transmits DR that start at like 5 or something).

Hmm.. it could be possibly done with a match and min. Except the AS923 regions which work somewhat opposite way and also depend on `DownlinkDwellTime`:
image

PR #357, opened 2025-03-04
Basic NewChannelReq stub implementation

plaes:
This adds basic stub for implementing NewChannelReq MAC command.
Should be good enough to commit as is, as a next step we'll need to implement following:
1. Define allowed frequency ranges for each region ( #359 )
2. Define supported frequencies for drivers (lora-phy)
3. Figure where to store the new channel information... current code has following:
   ```rust
   pub(crate) struct DynamicChannelPlan< const NUM_JOIN_CHANNELS: usize, R: DynamicChannelRegion<NUM_JOIN_CHANNELS>> {
      additional_channels: [Option<u32>; 5],   /// should I use this? But why only 5?
      channel_mask: ChannelMask<9>,
      last_tx_channel: u8,
      _fixed_channel_region: PhantomData<R>,
      rx1_offset: usize,
      rx2_dr: usize,
   }
   ```

LoRaWAN 1.0.4 documentation for NewChannelReq mentions:
>If the number of default channels is N (join channels?), the default channels go from 0 to N−1, and the acceptable range for ChIndex is N to 15. An
end-device SHALL be able to handle at least 16 different channel definitions.

But in the RP1.0.4:
> ...each region should support at least 24 channels (and maximum 80).

### DIFF
--- a/lorawan-device/Cargo.toml
+++ b/lorawan-device/Cargo.toml
@@ -41,6 +41,7 @@ lorawan = { path = "../lorawan-encoding", default-features = false, features = [
     "default-crypto",
     "defmt-03",
 ] }
+hex = "0.4.3"
 
 [features]
 default = ["all-regions", "class-c"]
--- a/lorawan-device/src/async_device/test/maccommands.rs
+++ b/lorawan-device/src/async_device/test/maccommands.rs
@@ -37,8 +37,40 @@ fn newchannelreq_invalid_eu868(
     // NewChannelReqPayload([0, 24, 79, 132, 80])
     // NewChannelReqPayload([1, 24, 79, 132, 80])
     // NewChannelReqPayload([2, 24, 79, 132, 80])
-    // In EU868 we are not allowed to modify 3 first channels
+    // EU868 - first 3 channels are join channels and readonly
     build_frm_payload(buf, "0700184f84500701184f84500702184f8450")
+}
+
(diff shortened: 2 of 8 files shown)


lthiery:
`   additional_channels: [Option<u32>; 5],   /// should I use this? But why only 5?`

just a note on this: this is an artifact of many concentrators being 8 channel. so the 3 base channel plus 5 additional. in reality, probably good to make it 64 channels?

lthiery (dismissed):
you are crushing

lthiery (approved):

plaes:
> ` additional_channels: [Option<u32>; 5], /// should I use this? But why only 5?`
> 
> just a note on this: this is an artifact of many concentrators being 8 channel. so the 3 base channel plus 5 additional. in reality, probably good to make it 64 channels?

Ok, follow up on this:
* A 16-channel gateway device will be used for all official Dynamic Channel Plan device certification testing
* A 64-channel gateway device will be used for all official Fixed Channel Plan device certification testing
* An 8-channel gateway device could be used for unofficial pre-testing conducted at LoRaWAN member labs. However, for official certification testing, 8-channel gateway devices will not be used. When using an 8-channel gateway, 125kHz Channels 0-7 and 500kHz Channel 64 must be configured on the gateway

So we need this to be either 16 or 64 channels depending on Dynamic/Region.

PR #362, opened 2025-03-10
adjust tx power for sx1262 in the lowest range

lucasgranberg:
There is a problem with setting the tx power. I did a test by looping over a range of `-9..=22` and giving a pulse at each power level. The result should be linear but looks like this
image

After this change it looks like this:
image

Still not completely linear but my device is not made for anything above 16dBm so the uneven result might be due to that.

I did also a test where I implemented this as suggested in the datasheet but the result was worse:
image
```rs
                match txp {
                    22 => {
                        self.set_pa_config(0x04, 0x07, DeviceSel::HighPowerPA).await?;
                        tx_params_power = 22;
                    }
                    20 => {
                        self.set_pa_config(0x03, 0x05, DeviceSel::HighPowerPA).await?;
                        tx_params_power = 22;
                    }
                    17 => {
                        self.set_pa_config(0x02, 0x03, DeviceSel::HighPowerPA).await?;
                        tx_params_power = 22;
                    }
                    _ => {
                        self.set_pa_config(0x02, 0x02, DeviceSel::HighPowerPA).await?;
                        tx_params_power = txp;
                    }
                }
```

### DIFF
--- a/lora-phy/src/sx126x/mod.rs
+++ b/lora-phy/src/sx126x/mod.rs
@@ -460,8 +460,10 @@ where
                     }
                     HIGH_POWER_MIN..=14 => {
                         self.set_pa_config(0x02, 0x02, DeviceSel::HighPowerPA).await?;
-                        // table indicates 14 dBm => txp = 22, therefore we add 8 to this range
-                        tx_params_power = txp as u8 + 8;
+                        // table indicates 14 dBm => txp = 22, therefore we should add 8 to this range
+                        // this however seems to be wrong when looking at the reference driver
+                        // https://github.com/STMicroelectronics/STM32CubeWL/blob/139e8d28bcec6af78dec8b52a9b9f9057868cc2e/Middlewares/Third_Party/SubGHz_Phy/stm32_radio_driver/radio_driver.c#L675
+                        tx_params_power = txp as u8;
                     }
                     _ => {
                         unreachable!("Invalid output power value for high power PA!")


CBJamo:
Do you have a standalone sx1262 to run the test with? It's possible the stm32wl has a different PA.

lthiery:
This reference might be interesting

https://github.com/Lora-net/SWL2001/blob/master/lbm_examples/radio_hal/ral_sx126x_bsp.c#L249-L315

I don't have time tonight to compare but I just wanted to drop the reference code. 

And if there's interest, this branch exists that builds the SWL2001 project: https://github.com/lora-rs/lora-rs/pull/233

lthiery:
And of course, appreciate the work and measurements you're doing!

lucasgranberg:
> Do you have a standalone sx1262 to run the test with? It's possible the stm32wl has a different PA.

no I don't and yes it is possible it is different. If it is then this code should be moved to the variant.

lucasgranberg:
> This reference might be interesting
> 
> https://github.com/Lora-net/SWL2001/blob/master/lbm_examples/radio_hal/ral_sx126x_bsp.c#L249-L315
> 
> I don't have time tonight to compare but I just wanted to drop the reference code. 
> 
> And if there's interest, this branch exists that builds the SWL2001 project: https://github.com/lora-rs/lora-rs/pull/233

yes. here they seem to skip all "optimal" values. doing that the curve looks like this:

sssilver:
[Would this be considered a "standalone SX1262"](https://www.cdebyte.com/products/E22-900M33S)? If so, I have a few, and am happy to perform the test.

lucasgranberg:
> [Would this be considered a "standalone SX1262"](https://www.cdebyte.com/products/E22-900M33S)? If so, I have a few, and am happy to perform the test.

yes that would be great, The code I used for testing looks like this:
```rs
#[cfg(feature = "continuous_wave")]
async fn continuous_wave(board: &mut Board) {
    use ::lorawan::device::Device as _;
    let mut lorawan_device = bsp::Board::get_lorawan_device(&mut board.lora).await;

    lorawan_device.radio().init().await.unwrap();
    let mdltn_params = lorawan_device
        .radio()
        .create_modulation_params(
            lora_modulation::SpreadingFactor::_12,
            lora_modulation::Bandwidth::_125KHz,
            lora_modulation::CodingRate::_4_5,
            869500000,
        )
        .unwrap();
    loop {
        for i in -9..=22 {
            let res = lorawan_device.radio().continuous_wave(&mdltn_params, i).await;
            embassy_time::Timer::after(Duration::from_millis(100)).await;
            lorawan_device.radio().enter_standby().await.unwrap();
            embassy_time::Timer::after(Duration::from_millis(100)).await;
            info!("{:?}", res);
        }
        lorawan_device.radio().enter_standby().await.unwrap();
        embassy_time::Timer::after(Duration::from_millis(5000)).await;
    }
}
```

plaes:
Ok, here's my "unscientific" testing with tools I have at hand.

nrf52840 + sx1262 (RAK4630):
image

plaes (approved):

lthiery (approved):

PR #366, opened 2025-03-18
Multicast: more fixes for MulticastGroupSetupDelete

lthiery:
When receiving a McGroupDelete msg for a group that is undefined, we should answer with the undefined bit and NOT include the multicast group id.

Furthermore, there was a bug in the encoding of the McGroupDeleteReqAns undefined bit.

### DIFF
--- a/lorawan-device/src/async_device/test/multicast.rs
+++ b/lorawan-device/src/async_device/test/multicast.rs
@@ -2,8 +2,12 @@ use super::*;
 use crate::async_device::McAddr;
 use lorawan::creator::DataPayloadCreator;
 use lorawan::keys::{McKEKey, McKey};
-use lorawan::multicast::{McGroupSetupAnsPayload, McGroupSetupReqCreator};
+use lorawan::multicast::{
+    parse_uplink_multicast_messages, McGroupDeleteReqCreator, McGroupSetupReqCreator,
+    UplinkRemoteSetup,
+};
 use lorawan::parser::{DataHeader, DataPayload, FRMPayload, PhyPayload};
+use std::time::Duration;
 
 fn handle_multicast_setup_req(
     _uplink: Option<Uplink>,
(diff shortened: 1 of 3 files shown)


plaes (dismissed):
No test? :)

lthiery:
> No test? :)

Will try to add tonight

ShakenCodes:
The HIL test cases in my codebase, which originally identified this issue, pass with these changes! (Not that there shouldn't be local tests added.)

lthiery:
I've added some tests but I feel like it needs a bit of work in the future so we can avoid the arbitrary delay while rxc_listen quietly does stuff in the background... but that's something to chew on for the future IMO

lorawan-device/src/async_device/test/multicast.rs
```diff
@@ -99,3 +120,136 @@ async fn test_multicast_remote_setup() {
...
+    })
+}
+
+fn handle_regular_downlink_msg<const FCNT: u32>(
+    _uplink: Option<Uplink>,
+    _config: RfConfig,
+    rx_buffer: &mut [u8],
+) -> usize {
+    let mut phy = DataPayloadCreator::new(rx_buffer).unwrap();
+    phy.set_f_port(1); // Remote multicast setup port
```
plaes, line 170: You sure about this comment?

lorawan-device/src/async_device/test/multicast.rs
```diff
@@ -34,32 +38,49 @@ fn handle_multicast_setup_req(
...
+fn verify_multicast_setup_ans(
+    uplink: Option<Uplink>,
+    _config: RfConfig,
+    _rx_buffer: &mut [u8],
+) -> usize {
+    verify_multicast_message(uplink, 200, |ans_data| {
+        let mut msgs = parse_uplink_multicast_messages(ans_data);
+        let msg = msgs.next().unwrap();
+        if let UplinkRemoteSetup::McGroupSetupAns(ans) = msg {
+            assert_eq!(ans.mc_group_id_header(), 0x01);
```
plaes, line 74: Why twice?

lorawan-device/src/async_device/test/multicast.rs
lthiery: copy pasta fail

lorawan-device/src/async_device/test/multicast.rs
```diff
@@ -34,32 +38,49 @@ fn handle_multicast_setup_req(
...
+            assert_eq!(ans.mc_group_id_header(), 0x01);
```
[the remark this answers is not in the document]
lthiery: non-sense from me

plaes (approved):

Issue #238, opened 2024-04-02
examples: Better documentation for stm32wl examples


plaes:
Apparently there are two official Nucleo devboards from STM (as of 2024.04) for stm32wl with different frequency ranges:
![image](https://github.com/lora-rs/lora-rs/assets/91895/ef10d102-1c79-4bc6-8ce2-d6e8cc70d92f)

Improve README for this example with a table for suggested frequencies for different regions these boards support.

plaes:
This came up in #369. Examples (or docs) should include list of things  to look for with pictures.

Things like RF switches and their cannot be be detected programmatically runtime and chips can be wired up differently to antennas and RF switches. Therefore README should provide some hints with pictures and/or schematics what to look for.

cc: @lucasgranberg

PR #369, opened 2025-03-24
Make use of all ctrl pins in stm32wl examples

lucasgranberg:
The example does not work for stm32wl-nucleo. It is not stated which board this example is for but the pins corresponds to the nucleo. The nucleo board needs to switch between rx and tx by alternating between two pins. Just using one does not work.

It was also not possible to use low_power_pa before so I added that to the switch too. Otherwise if somebody tries to use the low power PA without modifying the switch they will get in to trouble.

### DIFF
--- a/examples/stm32wl/src/bin/lora_lorawan.rs
+++ b/examples/stm32wl/src/bin/lora_lorawan.rs
@@ -55,23 +55,20 @@ async fn main(_spawner: Spawner) {
     }
     let p = embassy_stm32::init(config);
 
-    // Set CTRL1 and CTRL3 for high-power transmission, while CTRL2 acts as an RF switch between tx and rx
-    let _ctrl1 = Output::new(p.PC4.degrade(), Level::Low, Speed::High);
-    let ctrl2 = Output::new(p.PC5.degrade(), Level::High, Speed::High);
-    let _ctrl3 = Output::new(p.PC3.degrade(), Level::High, Speed::High);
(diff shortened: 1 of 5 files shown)


lucasgranberg:
I created this as a draft for now as I need to test this with real HW.

lucasgranberg:
This was tested on  NUCLEO-WL55JC2 (thanks @zeenix). This is but a quick fix and I plan to make it clearer that these examples are for the nucleo and also add an example for wio-e5 to show the difference between different PA setups. Ill do that in a separate PR.

plaes (approved):
I'm really confused because README already stated that these examples were for `NUCLEO-WL55JC1`  board...
I can speculate that any testing that was done with these was using two devices in close proximity (or near gateway.)

Ideally these ctrl1..ctrl3 variables could be renamed properly..

And see also #238 ;)

lucasgranberg:
> I'm really confused because README already stated that these examples were for NUCLEO-WL55JC1 board...

The readme states

> The NUCLEO-WL55JC1 board is a readily available development board for the STM32WL55JC SoC and it features an on-board debugger.

>Ideally these ctrl1..ctrl3 variables could be renamed properly..

Yeah they should be called that inside a variant that is called NUCLEO-WL55JC as they are nucleo specific. The names there now doesn't really reflect what is going on.

> And see also https://github.com/lora-rs/lora-rs/issues/238 ;)

Yes there could be a mention of it in the readme but as far as code goes I don't think there is anything we could query to see what RF end-stage  there is connected

PR #370, opened 2025-03-25
Migrate DR to lorawan::types

plaes:
My future intention is to add a new constructor to DataRateRange: `new_with_min_max()`. This will eventually used for in session to deal with NewChannelReq. But in order to support new need to store the dataraterange as well.
Currently we don't support adding new channels and as existing join channels are all using DR0..DR5 things just happen to work by chance.. :)

### DIFF
--- a/lorawan-encoding/src/types.rs
+++ b/lorawan-encoding/src/types.rs
@@ -143,6 +143,58 @@ impl<const N: usize> AsRef<[u8]> for ChannelMask<N> {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt-03", derive(defmt::Format))]
+#[repr(u8)]
+/// `DR` is a number from `0..=15` to indicate region-specific DataRate.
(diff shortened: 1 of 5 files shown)


lorawan-encoding/src/types.rs
```diff
@@ -143,6 +143,58 @@ impl<const N: usize> AsRef<[u8]> for ChannelMask<N> {
...
+            7 => DR::_7,
+            8 => DR::_8,
+            9 => DR::_9,
+            10 => DR::_10,
+            11 => DR::_11,
+            12 => DR::_12,
+            13 => DR::_13,
+            14 => DR::_14,
+            15 => DR::_15,
+            _ => unreachable!(),
```
lucasgranberg, line 192: maybe not related to this PR but should this give an Err instead of panic on unreachable?

lucasgranberg (approved):
nice clean-up, makes sense to move this

lorawan-encoding/src/types.rs
plaes: Now that you point it out, we should probably return Error::RFU?

PR #372, opened 2025-04-01
Add support for LinkADRReq (except nbtrans)

plaes:
This PR adds support for LinkADRReq MAC command (channel_mask, txpower and datarate), with the exception of `nbtrans`.

Tested with following regions:
* EU868 -  `TP_A_EU868_ED_MAC_104_BV_012_A`
* US915 - `TP_A_US915_ED_MAC_104_BV_012_A`

In the process found at least one nasty [infinite loop bug](https://github.com/lora-rs/lora-rs/pull/372/commits/cc07d2e08126798ca439105adab3911decd6b726) in logic where it attempts to randomly selected 500kHz channel...


lucasgranberg:
I used a peekable iterator for this
https://github.com/lucasgranberg/lorawan/blob/0c8446e037357f5949db7425c507948059080bae/src/mac/mod.rs#L415

That way I know what the next command is while going trough them

lucasgranberg:
Looking at my solution I think it was a good idea but I am not sure the answer is constructed correctly. I have never seen this from my network so I have not put that much effort in to it.

plaes:
> I used a peekable iterator for this https://github.com/lucasgranberg/lorawan/blob/0c8446e037357f5949db7425c507948059080bae/src/mac/mod.rs#L415
> 
> That way I know what the next command is while going trough them

Yeah, I'm going to do this as well. And for each command we need to validate channel mask:
* The channel mask enables a yet undefined channel
* the channel mask required all channels to be disabled
* channel mask is incompatible with the resulting data rate or TX power

This PR doesn't introduce any behavioral changes, it just refactors existing code to enabling smaller future changes (at least I hope this) .

lorawan-device/src/region/dynamic_channel_plans/mod.rs
```diff
@@ -144,42 +144,63 @@ impl<R: DynamicChannelRegion> RegionHandler for DynamicChannelPlan<R> {
...
-        &mut self,
+    fn channel_mask_get(&self) -> ChannelMask<9> {
+        self.channel_mask.clone()
+    }
+
+    fn channel_mask_set(&mut self, channel_mask: ChannelMask<9>) {
+        self.channel_mask = channel_mask;
+    }
+
+    fn channel_mask_update(
```
lucasgranberg, line 155: I don't understand what this fn does. what is mask, what is channel_mask. what is self in this context

plaes: Previously we had wrong behavior which updated existing channel plan (`self.channel_mask`) but there was no way to roll back changes in case the channel plan we ended up was erroneous.

Now, before this function gets called, we create a copy of existing channel plan and pass it into this function as `&mut mask`. Two other arguments come from LinkADRReq. Ideally we could rename the arguments in this way:
*  `mask (`&mut ChannelMask<9>` -> `channel_mask`
* `channel_mask_control` (`u8`) -> `control`
* `channel_mask` (`ChannelMask<2>` -> `mask`
And `&self` could be dropped as well.

plaes: Ok, updated the signature of this method. Although I'm not 100% satisfied with it yet...

lorawan-device/src/region/dynamic_channel_plans/mod.rs
```diff
@@ -144,40 +144,63 @@ impl<R: DynamicChannelRegion> RegionHandler for DynamicChannelPlan<R> {
...
+                    ch_mask.get_index(0) as u16 | ((ch_mask.get_index(1) as u16) << 8);
+                channel_mask.set_bank(0, ((ch_mask & 0b1) * 0xFF) as u8);
+                channel_mask.set_bank(1, ((ch_mask & 0b10) * 0xFF) as u8);
+                channel_mask.set_bank(2, ((ch_mask & 0b100) * 0xFF) as u8);
+                channel_mask.set_bank(3, ((ch_mask & 0b1000) * 0xFF) as u8);
+                channel_mask.set_bank(4, ((ch_mask & 0b10000) * 0xFF) as u8);
+                channel_mask.set_bank(5, ((ch_mask & 0b100000) * 0xFF) as u8);
+                channel_mask.set_bank(6, ((ch_mask & 0b1000000) * 0xFF) as u8);
+                channel_mask.set_bank(7, ((ch_mask & 0b10000000) * 0xFF) as u8);
+                channel_mask.set_bank(8, ((ch_mask & 0b100000000) * 0xFF) as u8);
```
lthiery, line 178: I know you didn't write it originally and you only modified it, but perhaps:
```suggestion
                for bank in 0..9 {
                    let bit_set = (ch_mask >> bank) & 1;
                    channel_mask.set_bank(bank, (bit_set * 0xFF) as u8);
                }
```

lthiery (dismissed):
Looks good to me. My only concerns are documented with TODO already

plaes:
@lthiery Please see the last commit...

lucasgranberg (approved):

lorawan-device/src/region/dynamic_channel_plans/mod.rs
plaes: Will be done as future improvement as there's also some controversy in the docs regarding `ch_mask_ctl == 5`  with follow up errata.

PR #374, opened 2025-04-04
Implement acknowledgment mechanism for RxParamSetupReq command

plaes:
RxParamSetupReq is another "edge-case" mac-command with its own special-case acknowledgment mechanism:
> The RXParamSetupAns command SHALL be used by the end-device to acknowledge the
receipt of a RXParamSetupReq command. The RXParamSetupAns command SHALL be
added in the FOpts field (if FPort is either missing or >0) or in the FRMPayload field (if
FPort=0) of all uplinks until a Class A downlink is received by the end-device. This
guarantees that, even in the case of an uplink frame loss, the Network is always aware of the
downlink parameters used by the end-device.

This PR implements the acknowledgment mechanism and adds basic testcase for it. All the other code including erroneously acking RX1 DR offset and RX2 DR have been left as they were and will be handled in upcoming patches.

~I'm somewhat unhappy with the resulting Uplink API though, we can end up multiple acknowledgment messages when someone decides to call `finalize()` multiple times...~

### DIFF
--- a/lorawan-device/src/mac/session.rs
+++ b/lorawan-device/src/mac/session.rs
@@ -129,6 +129,12 @@ impl Session {
         if let Ok(PhyPayload::Data(DataPayload::Encrypted(encrypted_data))) =
             lorawan_parse(rx.as_mut_for_read(), C::default())
         {
+            // If ignore_mac is false, we're dealing with Class A downlink and
+            // therefore can clear uplinks which need to be retained for acknowledgment
+            if !ignore_mac {
+                self.uplink.clear_mac_commands(false);
+            }
+
             #[cfg(feature = "certification")]
             if let Some(port) = encrypted_data.f_port() {
                 if port > 0 {
@@ -290,7 +296,7 @@ impl Session {
             &crypto_factory,
         ) {
             Ok(packet) => {
-                self.uplink.clear_mac_commands();
+                self.uplink.clear_mac_commands(true);
                 tx_buffer.clear();
                 tx_buffer.extend_from_slice(packet).unwrap();
             }
@@ -382,11 +388,20 @@ impl Session {
                         configuration.rx2_frequency = Some(freq);
                     }
 
+                    // RXParamSetupReq has its own acknowledgment mechanism, requiring
+                    // RXParamSetupAns with all uplinks until a Class A downlink is received
+                    // by the end-device.
                     let mut cmd = RXParamSetupAnsCreator::new();
                     cmd.set_rx1_data_rate_offset_ack(true)
                         .set_rx2_data_rate_ack(true)
                         .set_channel_ack(freq_ack);
+
                     self.uplink.add_mac_command(cmd);
+
+                    // TODO: An end-device that expects to receive Class C
+                    // downlink frames will send an uplink frame as soon
+                    // as possible after receiving a valid RXParamSetupReq
+                    // that modifies RX2 (Frequency or RX2DataRate fields).
                 }
                 RXTimingSetupReq(payload) => {
                     configuration.rx1_delay = super::del_to_delay_ms(payload.delay());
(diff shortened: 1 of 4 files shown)


lorawan-device/src/mac/session.rs
```diff
@@ -382,11 +386,20 @@ impl Session {
...
+                    // RXParamSetupReq has its own acknowledgment mechanism, requiring
+                    // RXParamSetupAns with all uplinks until a Class A downlink is received
+                    // by the end-device.
                     let mut cmd = RXParamSetupAnsCreator::new();
                     cmd.set_rx1_data_rate_offset_ack(true)
                         .set_rx2_data_rate_ack(true)
                         .set_channel_ack(freq_ack);
-                    self.uplink.add_mac_command(cmd);
+
+                    self.uplink.rxparam_pending_ack(cmd.ack_get());
```
lucasgranberg, line 397: there are also other commands that have the same ack mechanism. This is how I did it:
https://github.com/lucasgranberg/lorawan/blob/0c8446e037357f5949db7425c507948059080bae/src/mac/mod.rs#L824-L834

lorawan-encoding/src/maccommandcreator.rs
```diff
@@ -314,6 +314,17 @@ impl RXParamSetupAnsCreator {
...
         self
     }
+
+    /// Get acknowledgment value
+    pub fn ack_get(&self) -> u8 {
+        self.data[1]
+    }
+
+    /// Set acknowledgment value
+    pub fn ack_set(&mut self, ack: u8) -> &mut Self {
```
lucasgranberg, line 324: I am not a fan of  letting the user of this API inject any u8 in to here as all values of u8 are not valid.

lorawan-device/src/mac/session.rs
```diff
@@ -382,11 +386,20 @@ impl Session {
                         configuration.rx2_frequency = Some(freq);
                     }
 
+                    // RXParamSetupReq has its own acknowledgment mechanism, requiring
+                    // RXParamSetupAns with all uplinks until a Class A downlink is received
+                    // by the end-device.
                     let mut cmd = RXParamSetupAnsCreator::new();
                     cmd.set_rx1_data_rate_offset_ack(true)
```
lucasgranberg, line 393: are we always acking the dr offset? Is there no check for this?

lucasgranberg (changes requested):
Can we find a better solution to ack_set and ack_get in maccommandcreator?

lorawan-device/src/mac/session.rs
plaes: Yes, unfortunately this is how it has been from the "start".
I'll be working on these two once the acknowledgment mechanism is in place:
* RX2 DR - quite trivial to implement
* RX1 DR offset - somewhat more tricky as we have a lot more regions, but I have already an idea fleshed out based on feedback from #354

lorawan-encoding/src/maccommandcreator.rs
plaes: Yeah, I just needed something fast. Anyway, I think I can actually get rid of this API once I rework the uplink cleanup using same approach as you used.

lorawan-device/src/mac/session.rs
lucasgranberg: ok, step by step makes sense. I am ok with merging stuff that doesn't make it worse or break anything that was working

lorawan-encoding/src/maccommandcreator.rs
lucasgranberg: just save the whole RXParamSetupAnsCreator then. It is only one byte extra. I don't like putting "temporary" stuff in crates that are used outside of lora-rs.

lorawan-device/src/mac/session.rs
plaes: Thanks, stole it from you :)

lorawan-encoding/src/maccommandcreator.rs
plaes: I reworked the acknowledgment handling by retaining the mac commands and this allowed me to get rid of this API.

lucasgranberg (approved):

plaes:
> Can we find a better solution to ack_set and ack_get in maccommandcreator?

This addition was dropped.

PR #375, opened 2025-04-08
Set payload length correctly for SX127x Variant

UserJHansen:
Currently on the SX127x variant the payload length is set to 255 (0xff), causing PayloadSizeMismatch errors which this fixes

### DIFF
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -292,6 +292,10 @@ where
 
         C::set_packet_params(self, pkt_params).await?;
 
+        // Set the expected packet recieve size
+        self.write_register(Register::RegPayloadLength, pkt_params.payload_length)
+            .await?;
+
         // IQ inversion:
         // RegInvertiq - [0x33]
         // [6] - InvertIQRX
@@ -358,7 +362,6 @@ where
         self.write_register(Register::RegLna, lna_gain).await?;
 
         self.write_register(Register::RegFifoAddrPtr, 0x00u8).await?;
-        self.write_register(Register::RegPayloadLength, 0xffu8).await?;
 
         self.write_register(Register::RegOpMode, mode.value()).await
     }
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -292,9 +292,11 @@ where
 
         C::set_packet_params(self, pkt_params).await?;
 
-        // Set the expected packet receive size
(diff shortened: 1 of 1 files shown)


lora-phy/src/sx127x/mod.rs
```diff
@@ -292,6 +292,10 @@ where
 
         C::set_packet_params(self, pkt_params).await?;
 
+        // Set the expected packet recieve size
```
plaes, line 295: *receive

plaes (commented):
Do you have some example code / scenario about how this causes issues?

UserJHansen:
Yep, so for any situation where you don't have a buffer equal to 255 bytes it causes issues, e.g:
```rust
let recieve_buffer = [0; 11];
let rx_pkt_params = lora.create_rx_packet_params(
        4,
        true,
        recieve_buffer.len() as u8,
        true,
        false,
        &mdltn_params,
    ).unwrap();
lora.prepare_for_rx(lora_phy::RxMode::Continuous, &mdltn_params, &rx_pkt_params)
    .await
    .unwrap();
let packet = lora.rx(&rx_pkt_params, &mut recieve_buffer).await;
```

In this situation packet will always be a PayloadSizeMismatch(255, 11) error on the SX1276 radio.

lucasgranberg:
This makes sense to me but I have no working sx127x to test with. The manual says the default is 0x40 so idk how it has worked before. Has nobody received more than 64 bytes?

lucasgranberg:
Okay so set_payload does use this but this is only for tx
https://github.com/lora-rs/lora-rs/blob/9afed8a238d382aed979c68b0f73f0f5e8fd752a/lora-phy/src/sx127x/mod.rs#L327-L333

then we have this in do_rx (which this PR removes)
https://github.com/lora-rs/lora-rs/blob/9afed8a238d382aed979c68b0f73f0f5e8fd752a/lora-phy/src/sx127x/mod.rs#L360

I think this is safe to merge and would probably make the radio quit receiving faster on short messages without waiting for the timeout.

lucasgranberg:
Some other things I found while looking in to it more:
*Does this interfere with continuous_wave or tx mode as create_packet_params is called there too. I don't think it should but idk.
* Maybe we should only set it in implicit header mode as it is not used in explicit header mode?
* A "length of 0 is not permitted" so a check for that could be added at some point.

btw, 64 bytes was for FSK/OOK mode so nvm. for lora the default is 0x00 which is not permitted.

lucasgranberg (approved):
makes sense to me, I have not tested with hw but I believe this should be correct.

plaes (approved):

PR #376, opened 2025-04-08
Add experimental feature to lorawan-device

plaes:
Some time ago I half-implemened support for `RXParamSetupReq` MAC command, but there's still some work left. Therefore hide it behind `experimental` feature flag.


lucasgranberg (approved):
You know what you are doing with the maccommands and what works and what doesn't. At some point a debug message would be good so we can at least get some hint about a command not being implemented
https://github.com/lora-rs/lora-rs/blob/51bfe8329465bd9b2f8c40ed01053252f06a50bf/lorawan-device/src/mac/session.rs#L411

PR #377, opened 2025-04-08
Refactor process_irq_event

lonesometraveler:
This PR refactors the `process_irq_event` method to improve readability and maintainability by breaking it into smaller, more focused functions with clear responsibilities:
1. `process_irq_event`: Main function that orchestrates the IRQ handling flow
2. `get_irq_state`: Dedicated method for reading and interpreting IRQ flags
3. `clear_irq_status`: Handles clearing interrupts
4. `implicit_header_mode_handling` (for sx126x only): Isolates special case for single receive mode 

Changes
* Extracted IRQ status retrieval and interpretation logic
* Simplified the interrupt clearing by always clearing all interrupts
* Moved implicit header mode timeout behavior to a dedicated method 

The refactor preserves all existing functionality while improving the code’s overall clarity and maintainability.

### DIFF
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -494,22 +494,12 @@ where
         self.intf.iv.await_irq().await
     }
 
-    /// Process the radio IRQ. Log unexpected interrupts. Packets from other
-    /// devices can cause unexpected interrupts.
-    ///
-    /// NB! Do not await this future in a select branch as interrupting it
-    /// mid-flow could cause radio lock up.
-    async fn process_irq_event(
+    async fn get_irq_state(
         &mut self,
         radio_mode: RadioMode,
         cad_activity_detected: Option<&mut bool>,
-        clear_interrupts: bool,
     ) -> Result<Option<IrqState>, RadioError> {
         let irq_flags = self.read_register(Register::RegIrqFlags).await?;
-        if clear_interrupts {
-            self.write_register(Register::RegIrqFlags, 0xffu8).await?; // clear all interrupts
-        }
-
         match radio_mode {
             RadioMode::Transmit => {
                 if (irq_flags & IrqMask::TxDone.value()) == IrqMask::TxDone.value() {
@@ -553,6 +543,29 @@ where
         // If no specific IRQ condition is met, return None
         Ok(None)
     }
+
+    async fn clear_irq_status(&mut self) -> Result<(), RadioError> {
+        self.write_register(Register::RegIrqFlags, 0xffu8).await // clear all interrupts
+    }
+
+    /// Process the radio IRQ. Log unexpected interrupts. Packets from other
+    /// devices can cause unexpected interrupts.
+    ///
+    /// NB! Do not await this future in a select branch as interrupting it
+    /// mid-flow could cause radio lock up.
+    async fn process_irq_event(
+        &mut self,
+        radio_mode: RadioMode,
+        cad_activity_detected: Option<&mut bool>,
+        clear_interrupts: bool,
+    ) -> Result<Option<IrqState>, RadioError> {
+        let irq_state = self.get_irq_state(radio_mode, cad_activity_detected).await;
+        if clear_interrupts && irq_state.as_ref().is_ok_and(|state| state.is_some()) {
+            self.clear_irq_status().await?;
+        }
+        irq_state
+    }
+
     /// Set the LoRa chip into the TxContinuousWave mode
     async fn set_tx_continuous_wave_mode(&mut self) -> Result<(), RadioError> {
         C::set_tx_continuous_wave_mode(self).await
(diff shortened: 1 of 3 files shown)


lora-phy/src/sx127x/mod.rs
```diff
@@ -553,6 +543,29 @@ where
...
+    /// NB! Do not await this future in a select branch as interrupting it
+    /// mid-flow could cause radio lock up.
+    async fn process_irq_event(
+        &mut self,
+        radio_mode: RadioMode,
+        cad_activity_detected: Option<&mut bool>,
+        clear_interrupts: bool,
+    ) -> Result<Option<IrqState>, RadioError> {
+        let irq_state = self.get_irq_state(radio_mode, cad_activity_detected).await;
+        if clear_interrupts && irq_state.as_ref().is_ok_and(|state| state.is_some()) {
```
lucasgranberg, line 563: Before we cleared the irq flags if there were any unconditionally. If irq_state is an error ex. `Err(RadioError::TransmitTimeout)` would clear_irq_status trigger?.

lucasgranberg:
I think these changes makes the code easier to read. I just need to understand the logic changes.

lonesometraveler:
Thanks for your review. You’re right, I overlooked the timeout errors.

I believe the code below mirrors the behavior of the original: it doesn’t clear interrupts if the SPI read fails, but does clear them otherwise. What do you think?

```rust
async fn process_irq_event(
    &mut self,
    radio_mode: RadioMode,
    cad_activity_detected: Option<&mut bool>,
    clear_interrupts: bool,
) -> Result<Option<IrqState>, RadioError> {
    let irq_state = self.get_irq_state(radio_mode, cad_activity_detected).await;
    if clear_interrupts && (irq_state != Err(RadioError::SPI)) {
        self.clear_irq_status().await?;
    }
    irq_state
}
```

lucasgranberg:
I would just do it unconditionally. if we have an SPI error it would probably not work anyway. The error enum is kind of bloated and one could even ask if timeout is an error. That is why I think we should not consider it here when it comes to clearing the interrupts.

We can add it later if there is some actual issue with it.

lucasgranberg:
Another logic change here is this check: `irq_flags != 0`

I am not sure what will happen here. We have a bug on stm32wl where the irq line is spammed without any flags being set. This would make that problem even worse as there would be more SPI traffic.

lonesometraveler:
So, `irq_flags` could be `0` due to a bug on the STM32WL, and in that case, we don’t want to clear interrupts since doing so would require unnecessary SPI transactions? Did I understand your comment correctly?

If so, I think we could introduce another enum variant (unless we prefer using `Ok(None)` for that):

```rust
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
pub enum IrqState {
    /// Runs the loop until after the preamble has been received
    PreambleReceived,
    /// Runs the loop until the operation is fully complete
    Done,
    /// Represents a false interrupt (irq_flags == 0)
    FalseInterrupt,
}
```

This way, if it is a false interrupt, we can return early from `get_irq_state`.

```rust
let irq_flags = self.read_register(Register::RegIrqFlags).await?;
if irq_flags == 0 {
    return Ok(Some(IrqState::FalseInterrupt));
}
```

And call `clear_irq_status` only if it is not a false interrupt?

```rust
if clear_interrupts && irq_state != Ok(Some(IrqState::FalseInterrupt)) {
    self.clear_irq_status().await?;
}
```

lucasgranberg:
Do you have hw to test on? cleaning up and refactoring is good but making logic changes without testing on actual hw is not. Can we keep the scope of this PR to just refactoring without changing the logic?

I dont think it is a good idea to add stuff to mask a bug.

lonesometraveler:
I only have SX1262 devices, but I’ve been running this refactored code (plus some other additions) on RP2040s for a few weeks now without any issues.

I completely agree that we don’t want to change the logic here. My intention was to replicate the original behavior exactly.

So, in my view, between the two code snippets below, version A mirrors the original behavior, while version B clears interrupts unconditionally, even if the SPI read fails (which the original code does not do). Should we go with A?

**A**
```rust
async fn process_irq_event(
    &mut self,
    radio_mode: RadioMode,
    cad_activity_detected: Option<&mut bool>,
    clear_interrupts: bool,
) -> Result<Option<IrqState>, RadioError> {
    let irq_state = self.get_irq_state(radio_mode, cad_activity_detected).await;
    if clear_interrupts && (irq_state != Err(RadioError::SPI)) {
        self.clear_irq_status().await?;
    }
    irq_state
}
```

**B**
```rust
async fn process_irq_event(
    &mut self,
    radio_mode: RadioMode,
    cad_activity_detected: Option<&mut bool>,
    clear_interrupts: bool,
) -> Result<Option<IrqState>, RadioError> {
    let irq_state = self.get_irq_state(radio_mode, cad_activity_detected).await;
    if clear_interrupts {
        self.clear_irq_status().await?;
    }
    irq_state
}
```

lucasgranberg:
B for the sake of this PR. We can try to come up with a better way later.

lucasgranberg (approved):
I think this is good now. I think we have some bugs in when we actually trigger the implicit header mode but that is an issue for another day. I am not sure we actually reset the timer in all cases but I am not sure.

Thank you for your contribution @lonesometraveler

Issue #350, opened 2025-02-20
lora-phy: Document droppable interfaces


plaes:
Top-level docstrings are missing info that `process_irq_event` and its users cannot be used in `select` branch safely which can cause radio lockup.

```rust
radio.start_rx(..).await;
match select(radio.wait_for_irq, ...).await {}
radio.complete_rx().await;
```

lthiery:
Oh fair enough. But we do have a safe await for the IRQ itself, right?

plaes:
> Oh fair enough. But we do have a safe await for the IRQ itself, right?

Yeah, `wait_for_irq` should be safe.

PR #379, opened 2025-04-11
Add a function for post-IRQ RX data extraction

lonesometraveler:
This PR adds three new methods to support more flexible and decoupled RX handling:
	•`get_rx_result`: Extracts the received payload and packet status after a completed RX event (`IrqState::Done`).
	•`get_irq_state`: Returns the current IRQ state.
	•`clear_irq_status`: Clears the IRQ status.

These functions offer an alternative to the `complete_rx` flow, enabling external waiting on IRQs and deferring payload extraction until `IrqState::Done` is confirmed.  
Since `get_irq_state` can be safely dropped, this approach also avoids the issue described in #350, where `process_irq_event` should not be used in a `select` branch.

**Example usage:** 

```rust
match select(
    wait_for_rx_irq(&mut radio, &mod_params, &rx_params), // see below
    whatever,
)
.await
{
    Either::First(result) => {
        if let Err(e) = radio.clear_irq_status().await {
            error!("Error clearing IRQ: {}", e);
        }

        if result.is_ok() {
            let mut buf = [0u8; 256];
            match radio.get_rx_result(&rx_params, &mut buf).await {
                Ok((len, status)) => {
                    // Process received data
                }
                Err(e) => {
                    // Handle the error
                }
            }
        }
    }
    // ...
}

/// Starts RX and waits for an RX IRQ. This can be safely canceled.
async fn wait_for_rx_irq(
    radio: &mut LoraDevice,
    mod_params: &ModulationParams,
    rx_params: &PacketParams,
) -> Result<(), RadioError> {
    radio.prepare_for_rx(RxMode::Continuous, mod_params, rx_params)
        .await?;
    radio.start_rx().await?;

    loop {
        radio.wait_for_irq().await?;
        match radio.get_irq_state().await {
            Ok(Some(IrqState::Done)) => {
                return Ok(());
            }
            Ok(_) => yield_now().await,
            Err(e) => return Err(e),
        }
    }
}
```

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -95,6 +95,10 @@ where
     }
 
     /// Process an IRQ event and return the new state of the radio
+    ///
+    /// # Warning
+    /// This function is not safe to drop or cancel, as it calls `process_irq_event`, which must run to completion to avoid radio lockups.
+    /// Do not call this function within a select branch or in any context where it may be prematurely canceled.
     pub async fn process_irq_event(&mut self) -> Result<Option<IrqState>, RadioError> {
         self.radio_kind.process_irq_event(self.radio_mode, None, false).await
     }
@@ -220,6 +224,10 @@ where
     }
 
     /// Execute a transmit operation
+    ///
+    /// # Warning
+    /// This function is not safe to drop or cancel, as it calls `process_irq_event`, which must run to completion to avoid radio lockups.
+    /// Do not call this function within a select branch or in any context where it may be prematurely canceled.
     pub async fn tx(&mut self) -> Result<(), RadioError> {
         if let RadioMode::Transmit = self.radio_mode {
             self.radio_kind.do_tx().await?;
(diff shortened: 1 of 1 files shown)


CBJamo:
I like this change.

I'm inclined to also remove `wait_for_packet_start` here. It's not in a released version of the library (in fact, I'd bet I'm the only person using it), this is a better way to handle the same job, and not being cancel safe is a definite footgun. 

It'd also be great if all the functions that call `process_irq_state` were documented with the drop warning.

lonesometraveler:
@CBJamo Thanks so much for your review. I’m glad to hear you like the change.

Would you prefer that I remove `wait_for_packet_start` and add the drop warnings in this PR, or handle them in separate PRs? Personally, I think small, focused PRs are better, but I’m happy to include those changes here if you think it makes more sense.

plaes:
> @CBJamo Thanks so much for your review. I’m glad to hear you like the change.
> 
> Would you prefer that I remove `wait_for_packet_start` and add the drop warnings in this PR, or handle them in separate PRs? Personally, I think small, focused PRs are better, but I’m happy to include those changes here if you think it makes more sense.

You can include these changes within this PR.

lonesometraveler:
I’ve removed `wait_for_packet_start` and added the drop warnings. Let me know if I missed anything!

lora-phy/src/lib.rs
```diff
@@ -95,6 +95,10 @@ where
     }
 
     /// Process an IRQ event and return the new state of the radio
+    ///
+    /// # Warning
```
plaes, line 99: Maybe use `# Safety` instead of warning? As per https://std-dev-guide.rust-lang.org/policy/safety-comments.html

plaes: Actually, it doesn't matter. As both "Safety" and "Warning" look the same in docs.

plaes (approved):

plaes (changes requested):
Could you fix the commit titles to be more descriptive than "Update ..." ?

lonesometraveler:
Sorry about that. It was sloppy. I’ve updated the commit titles now.

plaes (approved):

PR #380, opened 2025-04-21
Fix doc comment in set_tx_power_and_ramp_time: remove unused tx_boosted_if_possible line

lonesometraveler:
The function comment for `set_tx_power_and_ramp_time` lists `tx_boosted_if_possible` as a parameter, but this parameter is not present in the function signature. This appears to be a leftover from an earlier version.

This PR removes the line to avoid confusion and keep the documentation accurate.

No functional changes.


plaes (approved):
Thanks!

PR #382, opened 2025-05-07
Remove crypto generic from lorawan-device

lucasgranberg:
This is kind of a ham-fisted way of removing the CryptoFactory generic from the lorawan device.

To achieve this the crypto factory doesn't live with the payloads anymore but is required to be passed in to any decode, validate_mic or any other function in lorawan-encoding that needs  crypto stuff.


lthiery (approved):

lthiery:
This is definitely what I would've done with a first pass. Thank you!

PR #385, opened 2025-05-21
DevStatusReq - wire in SNR support

plaes:
Adds support for SNR field in DevStatusReq/DevStatusAns MAC command.

And also updates README and reorganizes existing certification tests.

### DIFF
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -463,11 +463,14 @@ where
             )
             .await
             {
-                RxcWindowResponse::Rx(sz, _, timeout_fut) => {
+                RxcWindowResponse::Rx(sz, q, timeout_fut) => {
                     debug!("RXC window received {} bytes.", sz);
                     self.radio_buffer.set_pos(sz);
-                    let mac_response =
-                        self.mac.handle_rxc::<N, D>(&mut self.radio_buffer, &mut self.downlink)?;
+                    let mac_response = self.mac.handle_rxc::<N, D>(
+                        &mut self.radio_buffer,
+                        &mut self.downlink,
+                        q.snr(),
+                    )?;
                     match Self::handle_mac_response(
                         &mut self.radio_buffer,
                         &mut self.mac,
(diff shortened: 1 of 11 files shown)


lorawan-device/src/async_device/mod.rs
```diff
@@ -463,11 +463,14 @@ where
             )
             .await
             {
-                RxcWindowResponse::Rx(sz, _, timeout_fut) => {
+                RxcWindowResponse::Rx(sz, q, timeout_fut) => {
```
lucasgranberg, line 466: not a big fan of one letter variables

lorawan-device/src/mac/session.rs
```diff
@@ -314,21 +317,20 @@ impl Session {
...
-                    // TODO: Fill with proper values
-                    // - Battery: (255 - unable to measure, 1..254 - battery level, 0 - external power source)
-                    // - RadioStatus: (SNR: -32..31)
-                    // For now we just return dummy values
+                    // TODO: Battery information should come from device/application
+                    // Battery: (255 - unable to measure, 1..254 - battery level, 0 - external power source)
+                    // For now we just return dummy value of "255"
                     let mut cmd = DevStatusAnsCreator::new();
-                    let _ = cmd.set_battery(255).set_margin(0);
+                    let _ = cmd.set_battery(255).set_margin(snr);
```
lucasgranberg, line 333: Do we have any guarantees that the snr is 6-bit here? Seems to me like we get whatever the radio is returning, there is a cast from i16 to i8 here

lucasgranberg (commented):
Just a nit and a question, looks good otherwise

lorawan-device/src/mac/session.rs
lucasgranberg: https://github.com/lora-rs/lora-rs/blob/9223e421135ebf95af0f67a07ee33715c77465c7/lorawan-encoding/src/maccommandcreator.rs#L361-L368

Ok so there are some checks here but without using the result we probably end up with 0 here if the value is wrong. Idk if it ever would be a problem.

plaes: Yeah, I actually had these concerns as well, but I didn't want to go down the rabbit hole too deeply.
Ideally there should be a proper LoRaWAN "margin" type instead of i8.

lucasgranberg: true, should not be that big of a foot gun at this stage so good idea to not bloat the PR. All we get is bs answers but at least we get an answer. This command is used a lot by thingpark at least. Idk if it affects ADR but they like to show the battery status.

lucasgranberg (approved):

PR #386, opened 2025-05-25
NewChannelReq: Fix channel enabling

plaes:
Apparently when adding new channel, ChannelMask has to be updated as well.

Also, while at it, drop the `experimental` feature gating for `LinkADRReq` as it's mostly implemented.


lucasgranberg (approved):
Nice small change, good tests. Thanks!

Issue #383, opened 2025-05-19
LW 1.0.4 Certification for EU868 devices


plaes:
List of tests requires for EU868 Class A and Class C devices:
* [ ] 000 - Activation and Usage (Class C)
* [x] 000 - Activation and Usage
* [ ] 001A - Over the Air Activation - https://github.com/lora-rs/lora-rs/issues/335
* [x] 002 - Cryptography
* [x] 003 - Downlink Sequence Number
* [x] 004 - Confirmed Frames
* [ ] 005 - DevStatusReq MAC Command
* [ ] 006 - NewChannelReq MAC command for Dynamic Channel plan devices only
* [ ] 007 - DlChannelReq for Dynamic Channel plan devices only
* [ ] 008 - RXParameterSetupReq MAC Command
* [ ] 009 - RXTimingSetupReq MAC Command
* [x] 010 - TxParamSetupReq MAC Command 
* [ ] 011 - LinkCheckReq MAC Command
* [ ] 012A - LinkADRReq MAC Command (Part 1)
* [ ] 012B - LinkADRReq MAC Command (Part 2)
* [ ] 013 - DutyCycleReq MAC Command
* [ ] 014 - DeviceTimeReq MAC Command
* [ ] 015A - RX1 Window Test (Part 1)
* [ ] 015B - RX1 Window Test (Part 2)
* [ ] 016 - RX2 Receive Window Test
* [ ] 017 - RX1 and RX2 simultaneous frames
* [ ] 018 - RX Oversized Payload
* [ ] 019A - Maximum Allowed Payload (Part 1)
* [ ] 019B - Maximum Allowed Payload (Part 2)
* [ ] 020 - MAC command(s) in App-Payload and/or Frame Options
* [ ] 021 - Multiple MAC commands prioritization
* [ ] 022 - FPort 224 Deactivation

lucasgranberg:
This is huge! Such a good framework for getting stuff done 🤩

PR #387, opened 2025-05-26
Fix crash when filling MAC response buffer

plaes:
Apparently we can just truncate the MAC response payload when there's not enough room for all responses.

Issue was found thanks to LCTT testcase - "021 - Multiple MAC commands prioritization".


plaes:
Yes, current state of the MAC layer is not ideal, but at least one crash is fixed and it also passes the initial testcase (step 04):
image

> If the MAC command buffer is too large to fit in the frame, the DUT must truncate the buffer at
the end of the last MAC command that is able to fit within the frame.

There's still some other "horrors" I haven't yet figured out (see the final testcases at #383) and some of these rely on yet unimplemented MAC commands.

lucasgranberg (approved):

PR #388, opened 2025-05-27
Fix handling of implicit header on Sx127x radios.

bogdan2412:
When configuring an Sx127x radio to use implicit headers, the current behaviour is that we receive 255 bytes of data regardless of the payload length that we configured and the data appears to be garbage.

Instead of setting the payload length register to 255 in [do_rx], set it to the configured value in [set_packet_params].


lucasgranberg:
So this is a duplicate of https://github.com/lora-rs/lora-rs/pull/375

I have not been able to verify this bc I have not had the time to get my sx127x hardware up and running.
If you can confirm that #375 fixes the problem then we can go ahead and merge that one.

bogdan2412:
Oh, that's funny, I'll close this one and comment on the other.

PR #389, opened 2025-05-28
Add support for DlChannelReq MAC command

plaes:
Passes all LCTT tests.

### DIFF
--- a/lorawan-device/src/region/dynamic_channel_plans/mod.rs
+++ b/lorawan-device/src/region/dynamic_channel_plans/mod.rs
@@ -35,12 +35,25 @@ pub(crate) use in865::IN865;
 pub(crate) struct Channel {
     frequency: u32,
     _datarates: DataRateRange,
+    dl_frequency: Option<u32>,
 }
 
 impl Channel {
     /// Initialize Channel with frequency and supported minimum and maximum data rates
     fn new(f: u32, dr_min: DR, dr_max: DR) -> Self {
-        Self { frequency: f, _datarates: DataRateRange::new_range(dr_min, dr_max) }
+        Self::new_with_dr(f, DataRateRange::new_range(dr_min, dr_max))
+    }
+
+    fn new_with_dr(f: u32, dr: DataRateRange) -> Self {
+        Self { frequency: f, _datarates: dr, dl_frequency: None }
+    }
+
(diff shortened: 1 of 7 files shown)


lorawan-device/src/region/dynamic_channel_plans/mod.rs
```diff
@@ -35,12 +35,25 @@ pub(crate) use in865::IN865;
...
 }
 
 impl Channel {
     /// Initialize Channel with frequency and supported minimum and maximum data rates
     fn new(f: u32, dr_min: DR, dr_max: DR) -> Self {
-        Self { frequency: f, _datarates: DataRateRange::new_range(dr_min, dr_max) }
+        Self::new_with_dr(f, DataRateRange::new_range(dr_min, dr_max))
+    }
+
+    fn new_with_dr(f: u32, dr: DataRateRange) -> Self {
```
lucasgranberg, line 47: do we need two constructors here? Maybe it is enough to just have `new(f: u32, dr: DataRateRange)`

lorawan-device/src/region/fixed_channel_plans/mod.rs
```diff
@@ -263,6 +263,10 @@ impl<F: FixedChannelRegion> RegionHandler for FixedChannelPlan<F> {
         true
     }
 
+    fn channel_dl_update(&mut self, _: u8, _: u32) -> (bool, bool) {
```
lucasgranberg, line 266: idk if unreachable makes sense here. I went with returning `Err(Error::CommandNotImplementedForRegion)`. Not much better but at least gives some context.

The spec says "SHALL NOT be implemented" for fixed channel plan regions so I guess silent ignore is what we want and that we already get with `if region.has_fixed_channel_plan()`

lucasgranberg (approved):
My two comments are nothing major so Ill approve so you can choose how you want to proceed with those

lorawan-device/src/region/dynamic_channel_plans/mod.rs
plaes: I initially wanted to get rid of this as well, but the first variant is used a lot in tests.

lucasgranberg: Id vote for fixing the tests to keep the interface simple.

PR #392, opened 2025-06-03
Example for ESP32+RFM95 (SX1276) LoRaWAN

staugi:
Hello,

I would like to contribute a working and tested example of an ESP32 and RFM95 (SX1276) transceiver sending data using LoRaWAN.
I've also changed the directory structure of the existing esp32 example directory.

This is my first time contributing via a PR, so sorry in advance, If I'm not including/mentioning something I should've.

If there are problems with the code style, please do know, that the code style guidelines link in the contributing guidelines section is broken.


staugi:
Ok, it seems the examples didn't get built by the CI checks. I'm not sure how big of a problem this is.
If it is, is it in my power to fix?

lucasgranberg:
> Ok, it seems the examples didn't get built by the CI checks. I'm not sure how big of a problem this is. If it is, is it in my power to fix?

It just takes a while, seems to be fine now.

Thank you for your contribution! It may take a while until somebody can sit down and go trough this but hopefully it will be soon.
My knee jerk reaction is that this really is an esp32 + Sx127x example just with HopeRF naming. It is good that people see that it is possible to do with RFM95 but it might also be confusing idk. Did you use some ready made esp+RFM95 board or did you bread board it yourself?

staugi:
Yeah, I think in principle it should work with any sx127x modem.

The only catch with the RFM95 is that you must use the PA BOOST amp (tx_boost=true when creating the device config) as the RFO line is disconnected (I included this in the readme).

I did breadboard this from scratch. If it makes any difference I could make a wiring diagram or include photos.

lucasgranberg:
>The only catch with the RFM95 is that you must use the PA BOOST amp (tx_boost=true when creating the device config) as the RFO line is disconnected (I included this in the readme).

Ah I missed that part. Tricky, could be a separate variant but that would need some refactoring so probably good like this for now.

>I did breadboard this from scratch. If it makes any difference I could make a wiring diagram or include photos.

Photos are not needed, the schema can be gleaned from the code. Ofc a table in the readme with
esp pin numbering, function(ex. MOSI) and RFM95 pin numbering would be helpful for anybody wanting to replicate.

While reading the readme I saw
> ## This example works as of 2025-06-03 - no guaratees of working code provided after this date.

You could remove that as we are not giving any guarantees ever :D

staugi:
Let me know if you want me to make any other changes :)

lucasgranberg:
The table is good but could you add the RFM95 pin numbers too as they have them in the manual?
image

lucasgranberg:
Thank you!

Issue #82, opened 2022-09-08
Handle RXParamSetupReq


CBJamo:
While reviewing #81 and preparing a followup pr for IN865 and AS923 I found out that TTN uses a non-standard datarate for RX2 in the EU[1]. The TTN docs claim that the OTAA process should configure this, but I don't believe that's being handled at the moment. Missing RX2 windows probably isn't a huge issue (I don't think I've ever seen a RX2 downlink, though I'm not in the EU), and I think this can wait till someone has time to refactor the region code.

[1] https://www.thethingsnetwork.org/docs/lorawan/frequency-plans/

ivajloip:
Hi Caleb,

Thank you for bringing this to our attention! My guess is that they are using RX2 with SF9BW125 after the join and in that case indeed devices are supposed to handle that when using OTAA (it's an option in the accept message - I will have to check if that is currently implemented for the device). If they are using the same channel for the join request, then it doesn't sound aligned with my understanding of the specification, but it's been a while since I last read it :smile:

lthiery:
I believe this is simply related to handling `RXParamSetupReq` MAC Comand from the LNS (and the `RXParamSetupAns` response).

![image](https://github.com/ivajloip/rust-lorawan/assets/1606640/2d7d1072-bb50-4339-ac1d-fa8a53efda3c)

PR #395, opened 2025-06-05
RxParamSetupReq: Add RX1_DR_OFFSET support

### LINKED ISSUE #47: Support setting of a non-default RX2 frequency and DR
As documented in the description of `RXParamSetupAns`, the frequency for RX2 windows can be changed. This does not appear to be supported for some of the regional settings right now.

plaes:
Now that RX1_DR_OFFSET is implemented, RxParamSetupReq MAC command is fully implemented and `TP_A_EU868_ED_MAC_104_BV_008` test passes.

IN865 offset algorithm seems a bit gnarly though, but seems to be working.

Closes #47 and #82 (and supersedes #354)

### DIFF
--- a/lorawan-device/Cargo.toml
+++ b/lorawan-device/Cargo.toml
@@ -69,9 +69,7 @@ multicast = []
 ## Enable [`serde`](https://docs.rs/serde/latest/serde/) serialization/deserialization for data structures.
 serde = ["dep:serde", "lorawan/serde"]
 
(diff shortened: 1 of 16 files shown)


lucasgranberg (approved):
As far as I can see we have understood the specs the same way. There is a lot to parse here but should not break anything that works right now. The tests seems solid. in865 be crazy indeed. lgtm!

PR #400, opened 2025-06-11
Replace some trait methods with consts

plaes:
Clean up some region code by simplifying traits a bit...


lucasgranberg (approved):

lthiery (approved):
Nice clean up!

PR #401, opened 2025-06-27
Method that allows for faster frequency updating while receiving.

bogdan2412:
Add [rx_switch_channel] method which performs frequency updating conditional on the radio being in receive mode.

Previously, I believe this required the entire dance of creating modulation params, creating rx packet params and preparing for rx thereby resetting a lot of the same parameters over the SPI bus to unchanged values.

This does the simplest thing of:
- entering standby mode
- updating just the frequency registers
- resuming rx

On an ESP32 + Sx1276 device, I observed this reducing frequency updating times from 1815us to 490us on a release build (and 2710us to 1325us on a debug build). This is useful for applications involving continuous receiving coupled with periodic frequency updating, which could be quite timing sensitive.

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -279,7 +279,15 @@ where
         Ok(())
     }
 
-    /// Reconfigure radio that is presently receiving to listen to different frequency.
+    /// Reconfigure a radio that is presently receiving to listen on a different frequency,
+    /// without going through a full [`LoRa::prepare_for_rx`].
+    ///
+    /// Only the RF frequency is updated: modulation and packet parameters are kept, and
+    /// band-dependent configuration is not re-applied (image calibration, which the driver
+    /// performs once per [`LoRa::init`] at the first prepared frequency, and the sx127x
+    /// errata workarounds applied at [`LoRa::prepare_for_rx`] time). Intended for hops
+    /// between channels in the same band, like LoRaWAN channel hopping; for a band change,
+    /// use [`LoRa::prepare_for_rx`].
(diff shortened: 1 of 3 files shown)


lucasgranberg:
Another thing to check is how set_modulation_params works. There is a frequency_in_hz parameter in ModulationParams that would correspond with the new value. This is mostly a hack for an errata.

Also check if calibrate_image must be used. Seems important for sx126x

lthiery:
@bogdan2412 I used maintainer edit access to move this along (note your PR branch is your fork's `main`, so that ref was rebased):

1. Rebased onto current main.
2. Expanded the doc comment to state the limitation @lucasgranberg asked about: only the RF frequency is updated, so band-dependent configuration is not re-applied. The method is meant for same-band hops; band changes should go through `prepare_for_rx`.
3. Added an end-to-end test on the sx126x chip emulator we recently landed (#455-#458): RX on one frequency, switch, chip observed back in RX on the new PLL value, packet received through the switched-to session, and the not-receiving case rejected.

@lucasgranberg on your two checks:

- `calibrate_image`: the driver currently calibrates once per `init()` at the first prepared frequency and never again, so `prepare_for_rx` with a new frequency skips it too. This method is no worse than the existing path, and the doc now says same-band hops only. Re-calibrating on band change would be a fix for both paths and feels like a separate PR.
- `ModulationParams::frequency_in_hz`: it feeds band-dependent register values at `set_modulation_params` time (the sx127x errata workarounds, sx126x TX modulation quirk). Same story: stale only if you cross bands, which the doc now excludes.

On the precedent question: agreed a lower-level escape hatch is the cleaner long-term answer, but this shortcut is small, documented, and matches what LoRaWAN channel hopping needs today.

lthiery (approved):

PR #402, opened 2025-09-28
Fix continuous wave mode for SX1276

narottamroyal:
Enable TX switch and use LoRa registers by not setting bit 6 in `RegOpMode`.
Confirmed to work with an EBYTE 915M30S LoRa module:
image


lucasgranberg (approved):
I don't have the rubber stamps to merge this or the hw to test it but this makes sense to me. Idk how the wrong switch mode slipped trough. For any other reviewers, the regOpMode bit 6-5 change goes from 0b10(reserved) to 0b00(FSK)

PR #406, opened 2025-11-27
lora-phy: change ModulationParams & PacketParams field visibility

TheClams:
The goal is to allow implementation of the trait outside of the lora-rs crate.

### DIFF
--- a/lora-phy/src/mod_params.rs
+++ b/lora-phy/src/mod_params.rs
@@ -83,22 +83,32 @@ pub enum RxMode {
 }
 
 /// Modulation parameters for a send and/or receive communication channel
-#[allow(missing_docs)]
 pub struct ModulationParams {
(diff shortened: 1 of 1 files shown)


lucasgranberg (approved):
Simple enough, docs would be nice to have but can be added later as these are pretty self explanatory

TheClams:
> Simple enough, docs would be nice to have but can be added later as these are pretty self explanatory

Good point on the doc: since the members are public, it is better if it is documented.

TheClams:
> Thanks!
> 
> Are all these fields used in 2.4GHz as well? And how is frequency specified?
> 
> Right now, chips (and lorawan spec as well) use 100Hz units internally.

Yes everything is valid in 2.4GHz, and frequency can be provided directly in Hz (and like @lucasgranberg noted, this covers up to 4.2GHz).

PR #410, opened 2026-01-16
Feat/add lr1110

leftger:


lucasgranberg:
Feature gating does not help, the code is still in lora-rs and with limited resources to maintain it I do not think lora-rs should venture outside of lora stuff. The best approach imo would be to make a separate crate and make it implement RadioKind to make it compatible with lora-rs.

Issue #415, opened 2026-02-07
New example for the STM32WLE5Cx


jakobolason:
Hi, I've been working with the RAK3272s which has an RAK3172 which in turn has an STM32WLE5CC, which didn't require much altering to the STMWL example you had, but still took be -- an unexperienced user -- a week to get working. I have a working example which only alters the pins used and the internal clock to only use MSI.

I found out i needed some internal documents to get this working, as discussed here [in the RAK forums](https://forum.rakwireless.com/t/correct-pins-for-rak3272s-to-use-built-in-lora-radio/16722/4). 

Would a pull request with a working version for the RAK3272s into examples/ be of interest?

plaes:
It would be helpful. I actually have RAK3172 sitting unused, so I could then confirm Your findings

plaes:
stm32wle5cx example (#342) has been merged

PR #416, opened 2026-02-07
Stm32wle5cx examples

jakobolason:
Working examples for using the RAK3272s, as discussed per https://github.com/lora-rs/lora-rs/issues/415. It is mainly copied from the STM32WL example, but because I had some trouble getting it to work with the correct internal pins I thought it might be nice to have an extra example working.

The examples use MSI compared to the other examples, mainly because my board had weird behaviour when i tried to use the HSE that is in the RAK3172 ( I think there is a 32.768K external oscillator clock on that board, but the datasheets i could find do not actually describe it), and with my limited knowledge of embedded boards and what clocks are used for, I just used what worked for me :).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ exclude = [
     "examples/stm32l0",
     "examples/stm32wba",
     "examples/stm32wl",
+    "examples/stm32wle5cx",
 ]
 resolver = "2"
 
--- a/examples/stm32wle5cx/.cargo/config.toml
+++ b/examples/stm32wle5cx/.cargo/config.toml
@@ -1,0 +1,9 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+# replace your chip as listed in `probe-rs chip list`
+runner = "probe-rs run --chip STM32WLE5CC --speed 1000 --connect-under-reset"
+
+[build]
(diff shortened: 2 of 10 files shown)


plaes (approved):
Thanks.. Tested the lorawan join example with success!
```
27.565622 [DEBUG] RX2 received JoinSuccess (lorawan_device src/async_device/mod.rs:549)
27.565674 [INFO ] LoRaWAN network joined: JoinSuccess (lora_lorawan src/bin/lora_lorawan.rs:90)
27.565722 [INFO ] Joining LoRaWAN network (lora_lorawan src/bin/lora_lorawan.rs:72)
```

plaes:
@jakobolason If you have spare time, could you also borrow the code from [nrf52840 example](https://github.com/lora-rs/lora-rs/blob/597beac2a7774bd98874f1e5abfa9ac62eface81/examples/nrf52840/src/bin/lora_lorawan.rs#L27-L34) to override the EUI and APPKEY values from environment.

jakobolason:
Yes i could definitely do that, but practically since this is merged should i make a new PR?

plaes:
> Yes i could definitely do that, but practically since this is merged should i make a new PR?

Exactly! :)

Issue #418, opened 2026-02-13
Using LoRa.cad(...) in a loop makes the code panick?


jakobolason:
Hi I saw the cad and tried to experiment with some simple collision detection. It works fine when you just check LoRa.cad, but if i try to repeat it, it crashes.

I found that putting the cad detection in a loop:
```rust
loop {
    info!("Getting CAD ...");
    let mdltn_params = {
        match lora.create_modulation_params(
            SpreadingFactor::_10,
            Bandwidth::_250KHz,
            CodingRate::_4_8,
            LORA_FREQUENCY_IN_HZ,
        ) {
            Ok(mp) => mp,
            Err(err) => {
                info!("Radio error = {}", err);
                continue;
            }
        }
    };

    match lora.prepare_for_cad(&mdltn_params).await {
        Ok(()) => {}
        Err(err) => {
            info!("Radio error = {}", err);
            continue;
        }
    };

    match lora.cad(&mdltn_params).await {
        Ok(cad_activity_detected) => {
            if cad_activity_detected {
                info!("cad successful with activity detected")
            } else {
                info!("cad successful without activity detected")
            }
            Timer::after_secs(5).await;
        }
        Err(err) => info!("cad unsuccessful = {}", err),
    }
}
```

makes the code panic:
```
5.044665 [ERROR] panicked at /home/jakob/repos/pr-lora-rs/lora-phy/src/lib.rs:422:26:
internal error: entered unreachable code (panic_probe panic-probe-1.0.0/src/lib.rs:104)
Firmware exited unexpectedly: Multiple
 WARN probe_rs_debug::debug_info: No reliable return address is available, so we cannot determine the program counter to unwind the previous frame.
Core 0
    Frame 0: HardFault_ @ 0x08009828
       /home/jakob/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cortex-m-rt-0.7.5/src/lib.rs:1103:1
    Frame 1: HardFault <Cause: Escalated UsageFault <Cause: Undefined instruction>> @ 0x080049a2

    Frame 2: __udf @ 0x080049a2 inline
       /home/jakob/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cortex-m-0.7.7/src/../asm/inline.rs:181:5
    Frame 3: udf @ 0x00000000080049a2 inline
[...]

PR #419, opened 2026-02-13
phy: Don't clear IRQ status in CAD mode

plaes:
This allows us running CAD multiple times in sequence.

Fixes issue #418

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -409,7 +409,7 @@ where
             let mut cad_activity_detected = false;
             match self
                 .radio_kind
-                .process_irq_event(self.radio_mode, Some(&mut cad_activity_detected), true)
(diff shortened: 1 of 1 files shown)


lucasgranberg:
I don't understand why this solves the problem in https://github.com/lora-rs/lora-rs/issues/418.  What are we matching on when we hit the unreachable!() [here](https://github.com/plaes/lora-rs/blob/b4078f3d152355a118387d40bf3927952657a8d0/lora-phy/src/lib.rs#L422)? I wonder if the fix the unreachable match instead.

plaes:
Well, there's some kind of Gremlin either in our code or in smt32l hardware that makes things go sideways.
Continuous CAD loop works with regular sx1262 + nrf52840 variation (RAK4630 board).

lucasgranberg:
I make no sense 😅 I meant, should we not remove the unreachable instead?

PR #420, opened 2026-02-13
Refactor Channel Activity Detection result and add quirk for IRQ handling with SMT32 subghz devices

plaes:
Ok, this PR ballooned a bit and now includes all the changes in an intertwined mess:
1. Firstly, it refactors the mutable argument passing when dealing with CAD results
2. Secondly, it includes the sx127x bit flags handling
3. Thirdly, it adds a quirk for IRQ handling (currently only applied in CAD operation) - #418

### DIFF
--- a/lora-phy/src/mod_traits.rs
+++ b/lora-phy/src/mod_traits.rs
@@ -23,11 +23,11 @@ pub trait InterfaceVariant {
 /// Specifies an IRQ processing state to run the loop to
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum IrqState {
-    /// Runs the loop until after the preamble has been received
-    PreambleReceived,
-    /// Runs the loop until Channel Activity is Detected (or timeout occurs)
-    // TODO: Merge PreambleReceived + CadDetected into `Event`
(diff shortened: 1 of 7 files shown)


lora-phy/src/mod_traits.rs
```diff
@@ -23,8 +23,11 @@ pub trait InterfaceVariant {
 /// Specifies an IRQ processing state to run the loop to
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum IrqState {
-    /// Runs the loop until after the preamble has been received
-    PreambleReceived,
+    /// Runs until mode-specific event is detected.
+    /// Depending on Radio mode, it can be either:
+    /// * PreambleReceived in Rx mode
+    /// * CadDetected in CAD mode
+    Detect,
```
lucasgranberg, line 30: Could this be left as PreambleReceived  + CadDetected instead? Seems confusing to put them together

plaes: Yeah, unfortunately it is confusing. Anyway, I would have to revisit this patchset., because the first commit is actually buggy.

PR #422, opened 2026-02-16
phy: sx162x: Introduce common code for WriteRegister/ReadRegister calls

plaes:
SX162x code contains lot of "open-coded" WriteRegister/ReadRegister calls which made code somewhat hard to read:
* Introduce methods for single-byte value reads and writes
* Refactor `get_rx_payload` function to reduce nesting and get rid of top-level mutables.
* Additionally rehash workaround documentation based on latest available datasheet 

Also, I'm all open for better naming for `reg_r_8` and `reg_w_8` ;)

### DIFF
--- a/lora-phy/src/sx126x/mod.rs
+++ b/lora-phy/src/sx126x/mod.rs
@@ -137,14 +137,8 @@ where
             .await?;
 
         if symbol_num > 0 {
-            let val = exp + (mant << 3);
-            let buf = [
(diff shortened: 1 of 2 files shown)


plaes:
@lthiery @lucasgranberg

lucasgranberg (approved):
Makes the code cleaner. Maybe not the best names (as you said) for (reg_r/w_8) but doesn't really matter for an internal helper fn.

PR #427, opened 2026-04-24
phy: pre-command BUSY check per Semtech datasheets

swaits:
## Problem

While bringing up a downstream project (donglora), I hit a race where TX commands would occasionally fail. Tracked it down to lora-phy not following the BUSY-handshake algorithm Semtech recommends: the host should wait on BUSY *before* issuing each command, not after.

**Details:**

`SpiInterface::{write, write_with_payload, read,
read_with_status}` polls BUSY *after* each SPI op.

On fast MCUs (nRF52840 at 64 MHz, release build) this is racy — `embedded-hal-async`'s `wait_for_low` returns immediately if the pin is already low, and the chip takes up to T_SW ns (SX1261/2 §8.3, Fig. 8-3) to assert BUSY after NSS rises.

The host polls inside that pre-assertion window, sees a stale LOW, fires the next command while the chip is still busy, and the chip silently drops it. In my application, the first `lora.tx()` after a clean boot on Seeed Wio Tracker L1 hangs ~50% of the time.

## Authority

- [SX1261/2 datasheet, DS.SX1261-2.W.APP Rev. 1.2](https://cdn.sparkfun.com/assets/6/b/5/1/4/SX1262_datasheet.pdf) §8.3.1
  "BUSY Control Line": *"it is essential to wait for the BUSY line to
  go low before sending an SPI command."*
- [LR1121 User Manual, UM.LR1121.W.APP Rev. 1.2](https://files.waveshare.com/wiki/Core1121/UserManual_LR1121_v1_2.pdf) §3
  "Host-Controller Interface": *"it is necessary to check the status
  of BUSY prior to sending a command."* (Applies to the LR11xx family
  — see also [LR1110 DS Rev. 1.1](https://download.mikroe.com/documents/datasheets/LR1110_datasheet.pdf) §1.2.5.)
- [RadioLib](https://github.com/jgromes/RadioLib/blob/cffd5af/src/Module.cpp)
  `Module::SPIwriteStream` / `SPIreadStream` call `waitForBusy()` at
  the start of every transaction.
- The [SX1276/77/78/79 datasheet](https://cdn.sparkfun.com/assets/7/7/3/2/2/SX1276_Datasheet.pdf)
  Table 2 exposes no BUSY pin, so
  `GenericSx127xInterfaceVariant::wait_on_busy` stays a no-op and the
  change is behaviorally identical for SX127x.

## Fix

Replace the post-command BUSY wait with a pre-command BUSY check at the start of each `SpiInterface` / `Lr1110SpiInterface` read/write method. Drop the redundant explicit `wait_on_busy()` calls in `sx126x` (after Calibrate, `ensure_ready` non-sleep arm) and `lr1110::mod` (4 regmem/bootloader sites) — all now covered by the interface pre-wait.

The `is_sleep_command` flag keeps its name but now gates the pre-wait; set `true` only when BUSY is held HIGH by design (the wake-from-sleep GetStatus pulse in `sx126x::ensure_ready`). Two callsite flips as a result: `ensure_ready` Sleep arm `false`→`true`; `set_sleep` `true`→`false`.

No public API signature changes.

## Test

- Hardware: Seeed Wio Tracker L1 (nRF52840 + SX1262), 20 consecutive
  clean-boot TX cycles → 20/20 (vs ~50% before).

### DIFF
--- a/lora-phy/src/interface.rs
+++ b/lora-phy/src/interface.rs
@@ -17,45 +17,42 @@ where
         Self { spi, iv }
     }
 
-    // Write a buffer to the radio.
+    // Write a buffer to the radio. Pre-waits for BUSY low per SX1261/2 §8.3.1.
+    // `is_sleep_command` skips the pre-wait for the wake-from-sleep pulse where
+    // BUSY is held HIGH until NSS falls (see `sx126x::ensure_ready`).
     pub async fn write(&mut self, write_buffer: &[u8], is_sleep_command: bool) -> Result<(), RadioError> {
-        self.spi.write(write_buffer).await.map_err(|_| SPI)?;
-        trace!("write: {=[u8]:02x}", write_buffer);
-
         if !is_sleep_command {
             self.iv.wait_on_busy().await?;
         }
+        self.spi.write(write_buffer).await.map_err(|_| SPI)?;
+        trace!("write: {=[u8]:02x}", write_buffer);
 
         Ok(())
     }
 
-    // Write
+    // Write a command buffer followed by an inline payload. Pre-waits as `write()`.
     pub async fn write_with_payload(
         &mut self,
         write_buffer: &[u8],
         payload: &[u8],
         is_sleep_command: bool,
     ) -> Result<(), RadioError> {
+        if !is_sleep_command {
+            self.iv.wait_on_busy().await?;
+        }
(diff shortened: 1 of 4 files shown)


swaits:
FYI, after more testing, dialed the diff back to be significantly more conservative and reopened here: https://github.com/lora-rs/lora-rs/pull/428

PR #428, opened 2026-04-26
phy: pre-command BUSY check before each SPI op

swaits:
## Problem

`SpiInterface` / `Lr1110SpiInterface` wait on BUSY only _after_ each SPI op. SX1261/2 DS §8.3.1 and the LR1121 user manual §3 require waiting for BUSY low _before_ each command too — otherwise, on fast MCUs, the next command can race into the chip during the BUSY-assertion window and be silently dropped.

## Fix

Add `check_device_ready()` — a pre-command BUSY wait — at the start of each read/write method, bracketing each op with the existing post-op `wait_on_busy()`. `is_sleep_command` gates both waits; `ensure_ready`'s wake-from-sleep `GetStatus` pulse passes `is_sleep_command=true` so the pre-wait can't deadlock on the asleep chip.

SX127x has no BUSY pin — `wait_on_busy` is a no-op there, so this is a no-op for SX127x. No public API changes.

Prior art for the pre-command wait: LoRaMac-node, STM32WL HAL, RadioLib, TinyGo sx126x.

## Test

- Hardware: Seeed Wio Tracker L1 (nRF52840 + SX1262), 20 consecutive clean-boot TX cycles → 20/20 (vs ~50% before).

### DIFF
--- a/lora-phy/src/sx126x/mod.rs
+++ b/lora-phy/src/sx126x/mod.rs
@@ -339,8 +339,11 @@ where
     async fn ensure_ready(&mut self, mode: RadioMode) -> Result<(), RadioError> {
         match mode {
             RadioMode::Sleep | RadioMode::Receive(RxMode::DutyCycle(_)) => {
+                // Wake-from-sleep pulse: pass is_sleep_command=true to skip both
+                // BUSY waits, which would otherwise deadlock on the asleep chip.
+                // The next SPI op's pre-command wait absorbs the wake latency.
                 let op_code_and_null = [OpCode::GetStatus.value(), 0x00u8];
-                self.intf.write(&op_code_and_null, false).await?;
+                self.intf.write(&op_code_and_null, true).await?;
             }
             _ => self.intf.iv.wait_on_busy().await?,
         }
--- a/lora-phy/src/interface.rs
+++ b/lora-phy/src/interface.rs
@@ -17,8 +17,16 @@ where
         Self { spi, iv }
     }
 
+    // Wait for BUSY low before issuing a command (SX1261/2 DS §8.3.1).
+    async fn check_device_ready(&mut self) -> Result<(), RadioError> {
+        self.iv.wait_on_busy().await
+    }
+
     // Write a buffer to the radio.
     pub async fn write(&mut self, write_buffer: &[u8], is_sleep_command: bool) -> Result<(), RadioError> {
+        if !is_sleep_command {
+            self.check_device_ready().await?;
+        }
         self.spi.write(write_buffer).await.map_err(|_| SPI)?;
         trace!("write: {=[u8]:02x}", write_buffer);
 
@@ -36,6 +44,9 @@ where
         payload: &[u8],
         is_sleep_command: bool,
     ) -> Result<(), RadioError> {
+        if !is_sleep_command {
+            self.check_device_ready().await?;
+        }
         let mut ops = [Operation::Write(write_buffer), Operation::Write(payload)];
         self.spi.transaction(&mut ops).await.map_err(|_| SPI)?;
         trace!("write_buf: {=[u8]:02x} -> {=[u8]:02x}", write_buffer, payload);
@@ -49,6 +60,7 @@ where
 
     // Request a read, filling the provided buffer.
     pub async fn read(&mut self, write_buffer: &[u8], read_buffer: &mut [u8]) -> Result<(), RadioError> {
+        self.check_device_ready().await?;
         {
             let mut ops = [Operation::Write(write_buffer), Operation::Read(read_buffer)];
 
@@ -69,6 +81,7 @@ where
 
     // Request a read with status, filling the provided buffer and returning the status.
     pub async fn read_with_status(&mut self, write_buffer: &[u8], read_buffer: &mut [u8]) -> Result<u8, RadioError> {
+        self.check_device_ready().await?;
         let mut status = [0u8];
         {
             let mut ops = [
--- a/lora-phy/src/lr1110_interface.rs
+++ b/lora-phy/src/lr1110_interface.rs
@@ -29,8 +29,16 @@ where
         Self { spi, iv }
     }
 
+    // Wait for BUSY low before issuing a command (LR1121 UM §3).
+    async fn check_device_ready(&mut self) -> Result<(), RadioError> {
+        self.iv.wait_on_busy().await
+    }
+
     // Write a buffer to the radio.
     pub async fn write(&mut self, write_buffer: &[u8], is_sleep_command: bool) -> Result<(), RadioError> {
+        if !is_sleep_command {
+            self.check_device_ready().await?;
+        }
         self.spi.write(write_buffer).await.map_err(|_| SPI)?;
         trace!("write: {=[u8]:02x}", write_buffer);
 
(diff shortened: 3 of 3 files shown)


swaits:
Update:

I've now soaked these changes a pretty good amount of time across a variety of different hardware.

I can't say with 100% certainty, of course, but I'm getting confident this improves stability by following the Semtech application notes more closely.

What's the plan to test, validate, and eventually merge?

lucasgranberg:
Thanks for taking the time to look in to this. Yes we will try to validate this and then merge.

lora-phy/src/sx126x/mod.rs
```diff
@@ -716,6 +727,15 @@ where
 
         self.reg_w_8(Register::RxGain, rx_gain_final).await?;
 
+        // Clear stale IRQ flags (CADDone / CADDetected / RxDone) BEFORE
+        // SetCadParams/SetCAD. Without this, a previously-asserted IRQ
+        // bit can keep DIO1 from re-edging when the new CAD completes,
+        // and `wait_for_irq` either returns garbage or hangs. Required
+        // by the Semtech CAD/LBT guidebook §8.2 step 9 and rule of
```
lucasgranberg: Which application note is this? I can't find the document

swaits:
You're right — that citation is bogus. There's no "CAD/LBT guidebook" §8.2; apologies for letting slop slip in. I try to be detailed and meticulous, but missed that one.

The real anchor is SX1261/2 datasheet §8.5 ("IRQ Handling") plus the ClearIrqStatus / SetDioIrqParams command descriptions (§13.3.4 / §13.3.1): IRQ status bits are latched, and DIO1 is the OR of masked status bits, so it only re-edges when a previously-zero masked bit goes high.

Honest provenance: I hit this developing DongLoRa, and pre-clearing the IRQ unwedged it. That's empirical, not spec-mandated — Semtech's own `RadioStartCad()` in LoRaMac-node doesn't pre-clear. I can't guarantee it's the right fix versus a symptom of something else in my code. Happy to drop, narrow, or keep as defensive bracketing — your call.

Meanwhile I pushed an update to the docstring.

Datasheet (Sparkfun mirror); https://cdn.sparkfun.com/assets/6/b/5/1/4/SX1262_datasheet.pdf

swaits:
FWIW, this isn't novel. I looked at other implementations to see what they do. Check these out, please:

**Datasheet SX1261/2 §8.3.1:** *"it is essential to wait for the BUSY line to go low before sending an SPI command."*

- **Semtech LoRaMac-node** — `SX126xWriteCommand` calls `SX126xCheckDeviceReady()` *before* NSS drops, then `SX126xWaitOnBusy()` after. Same pattern in every other command function in the file. → https://github.com/Lora-net/LoRaMac-node/blob/master/src/boards/NucleoL476/sx1262mbxcas-board.c#L171-L298

- **STM32WL HAL (ST official)** — `HAL_SUBGHZ_ExecSetCmd` calls `SUBGHZ_CheckDeviceReady()` *before* the SPI transmit (L976) and `SUBGHZ_WaitOnBusy()` after (L1002). Pre **and** post. → https://github.com/STMicroelectronics/stm32wlxx-hal-driver/blob/main/Src/stm32wlxx_hal_subghz.c#L960-L1025

- **ARM Mbed (Semtech-authored driver)** — `write_opmode_command` does `while (_busy) { /* do nothing */ }` after dropping NSS and *before* `_spi.write(cmd)`. → https://github.com/ARMmbed/mbed-semtech-lora-rf-drivers/blob/master/SX126X/SX126X_LoRaRadio.cpp#L563-L577

- **RadioLib (jgromes)** — `Module::SPItransferStream` has an explicit `// ensure GPIO is low` block that spins on `digitalRead(gpioPin)` *before* sending the SPI command. → https://github.com/jgromes/RadioLib/blob/master/src/Module.cpp#L356-L370

- **TinyGo sx126x driver** — `ExecSetCommand` calls `CheckDeviceReady()` (which calls `WaitWhileBusy()`) *before* dropping NSS. Same in `ExecGetCommand` and `ReadRegister`. → https://github.com/tinygo-org/drivers/blob/release/sx126x/sx126x.go#L499-L516

Post-only is the unusual choice. On fast MCUs (nRF52840 @ 64 MHz) that's exactly the race §8.3.1 warns about.

---

**On the CAD IRQ pre-clear (the empirical bit):** RadioLib's `startChannelScan` runs the same sequence the PR now does — `standby()` → `setDioIrqParams()` → `clearIrqStatus()` → `setCad()`:

→ https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L591-L614

LoRaMac-node's `RadioStartCad()` skips the pre-clear, but it runs under a single owner with strict mode tracking. lora-rs is a shared driver where the app may have its own IRQ history, so defensive pre-clear matches RadioLib's posture for the same reason.

lora-phy/src/sx127x/mod.rs
```diff
@@ -438,6 +438,16 @@ where
         }
         self.write_register(Register::RegLna, lna_gain_final).await?;
 
+        // Clear stale CAD IRQ flags before entering CAD mode. Without
+        // this, a previously-asserted CadDone bit can keep DIO mapped
+        // to "active" and the host-side IRQ wait may not see a fresh
+        // edge. Required by SX1276/77/78/79 datasheet's documented CAD
+        // flow ("clear CadDone/CadDetected IRQs → set mode = CAD →
+        // wait for CadDone IRQ") and the Semtech CAD/LBT guidebook
```
lucasgranberg: still references to something that doesn't exist.

swaits: Ack. I'll clean that up and hope that's the last one.

lucasgranberg:
I still don't get it, you write about check_device_ready but your code has clear_irq_status and set_standby. Would it not be clearer to have a check_device_ready method if it doesn't exist?

Also do you feel your comment style in the code matches the rest of the project?

Anyway I can't really review or test this as I don't have the hardware for it. I run out of brain tokens trying to wade trough all the text and links and comments you have added for essentially 3 lines of code. It would be much easier to review if there was less slop. If you would just read your own code before submitting it would also help.

swaits:
Let me clean it up and resubmit. 

To be clear, I read the code very carefully. I've tested it extensively. I refined the code to its simplest form over many iterations and much trial and error, and deep debugging sessions. 

I used AI to help back the code changes with solid reasoning and that is where the slop slipped through. 

The code is legit. Let me get the test cleaned up for you. Thanks!

swaits:
@lucasgranberg -- note that I'm in process of testing this. Please don't look at it until I ping you here as it passing all my testing, again.

swaits:
Ok. As requested:

- simplified commentary significantly
- matched surrounding code style
- extracted to helper `fn` (I avoided earlier to create a minimal change, but I like this)

Fully tested on my end and seems to maintain the same stability the previous version showed.

darconeous:
I'm bringing up [UMSH](https://github.com/darconeous/umsh/) on the [Seeed SenseCap T1000-E](https://www.seeedstudio.com/SenseCAP-Card-Tracker-T1000-E-for-Meshtastic-p-5913.html), which has a LN111x.

I'll verify this change and likely have a few others shortly.

swaits:
Thanks @darconeous.

BTW, UMSH looks cool. I may try to build out a DongLoRa adapter for it which would give you some support for a few more boards to experiment with.

PR #432, opened 2026-04-30
phy: sx127x: clear stale IRQ flags before mode changes

swaits:
## Problem

`RegIrqFlags` is W1C with no auto-clear on `RegOpMode` transitions (SX1276 DS §4.1.2.4). A stale flag keeps DIO0 latched "active" across a mode change, so the host's edge-triggered `wait_for_irq` fires early or misses the next real edge.

## Fix

- `set_irq_params`: move the `RegIrqFlags = 0xFF` clear from the end of the function to the start — cleared before the DIO remap, never adjacent to a caller's `RegOpMode` write.
- `do_rx`: clear flags before entering Rx — covers the `listen()` path, which never calls `set_irq_params`.

Matches RadioLib's `stageMode` discipline (CLEAR → CONFIG → MODE_CHANGE). No public API changes.

Consolidates #431 (the `do_rx` clear) with this PR's `set_irq_params` reorder; #431 is closed in favour of this.

## Test

- Hardware: LilyGo T-Beam classic (ESP32 + SX1276), donglora downstream firmware. Pre-fix (with the companion `do_rx` clear applied but the trailing-clear ordering preserved): sustained TX-spam workload wedges the chip within seconds — RETRY storms fill the host log, the OLED reverts to splash mode, ai-bot eventually disconnects. Post-fix: same workload runs indefinitely without wedge across many minutes of sustained traffic.
- SX1262 boards (Heltec V3 USB, LilyGo T-Beam Supreme) verified compile-clean — change only touches `sx127x/mod.rs`.

### DIFF
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -382,9 +382,10 @@ where
 
         self.write_register(Register::RegFifoAddrPtr, 0x00u8).await?;
 
-        // Clear stale IRQ flags before entering Rx. RegIrqFlags is W1C with no
-        // auto-clear on mode change (SX1276 DS §4.1.2.4); covers the listen()
-        // path, which does not call set_irq_params.
+        // Interrupt flags stay latched until the host clears them by writing a 1
+        // (SX1276 DS §4.1.2.4); entering Rx does not reset them. Clear here so a
+        // flag left over from an earlier operation can't read as a result of this
+        // one; this also covers listen(), which never calls set_irq_params.
         self.clear_irq_status().await?;
 
         self.write_register(Register::RegOpMode, mode.value()).await
@@ -461,10 +462,12 @@ where
     // enable interrupts on DIO pins (sx127x has multiple),
     // and allow interrupts.
     async fn set_irq_params(&mut self, radio_mode: Option<RadioMode>) -> Result<(), RadioError> {
-        // Clear stale IRQ flags before remapping DIO. RegIrqFlags is W1C with no
-        // auto-clear on mode change (SX1276 DS §4.1.2.4); clearing first keeps
-        // it away from a caller's adjacent RegOpMode write.
-        self.write_register(Register::RegIrqFlags, 0xffu8).await?;
+        // Interrupt flags stay latched until the host clears them by writing a 1
+        // (SX1276 DS §4.1.2.4); mode changes do not reset them. Clear before the
+        // DIO remap so a leftover flag can't sit on a freshly mapped DIO line,
+        // where the MCU's edge-triggered interrupt would either fire on the stale
+        // flag or never see an edge for the next real one.
+        self.clear_irq_status().await?;
 
         match radio_mode {
             Some(RadioMode::Transmit) => {
--- a/lora-phy/src/sx127x/mod.rs
+++ b/lora-phy/src/sx127x/mod.rs
@@ -382,6 +382,11 @@ where
 
         self.write_register(Register::RegFifoAddrPtr, 0x00u8).await?;
 
+        // Clear stale IRQ flags before entering Rx. RegIrqFlags is W1C with no
+        // auto-clear on mode change (SX1276 DS §4.1.2.4); covers the listen()
+        // path, which does not call set_irq_params.
+        self.clear_irq_status().await?;
+
         self.write_register(Register::RegOpMode, mode.value()).await
     }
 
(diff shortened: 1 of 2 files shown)


lthiery (commented):
This looks correct to me but I haven't been deep in the radio libs. @lucasgranberg ?

lucasgranberg:
I don't have the hw to test this so I am not really the right one to approve or do anything about sx1276 stuff.

CLEAR->CONFIG->MODE_CHANGE makes sense to me.

The comments about sx1276DS 4.1.2.4 I don't really understand. That section says:
>4.1.2.4. Interrupts in LoRa Mode
Two registers are used to control the IRQ in LoRa mode, the register RegIrqFlagsMask which is used to mask the
interrupts and the register RegIrqFlags which indicates which IRQ has been trigged.
In the register RegIrqFlagsMask, setting a bit to ‘1’ will mask the interrupt, meaning this interrupt is disactivated. By default
all the interrupt are available.
In the register RegIrqFlags, a ‘1’ indicates a given IRQ has been trigged and then the IRQ must be clear by writing a ‘1’. 

So yes it is cleared by writing a 1 but that is pretty clear already.

Instead of using `self.write_register(Register::RegIrqFlags, 0xffu8).await?;`as before we could probably just use `self.clear_irq_status().await?;`. That way it is clear what happens even with less comments.

Also what does W1C mean?

lora-phy/src/sx127x/mod.rs
```diff
@@ -445,6 +450,11 @@ where
     // enable interrupts on DIO pins (sx127x has multiple),
     // and allow interrupts.
     async fn set_irq_params(&mut self, radio_mode: Option<RadioMode>) -> Result<(), RadioError> {
+        // Clear stale IRQ flags before remapping DIO. RegIrqFlags is W1C with no
+        // auto-clear on mode change (SX1276 DS §4.1.2.4); clearing first keeps
+        // it away from a caller's adjacent RegOpMode write.
+        self.write_register(Register::RegIrqFlags, 0xffu8).await?;
```
lthiery:
```suggestion
        // Interrupt flags stay latched until the host clears them by writing a 1
        // (SX1276 DS §4.1.2.4) — mode changes do not reset them. Clear before the
        // DIO remap so a leftover flag can't sit on a freshly mapped DIO line,
        // where the MCU's edge-triggered interrupt would either fire on the stale
        // flag or never see an edge for the next real one.
        self.clear_irq_status().await?;
```

lora-phy/src/sx127x/mod.rs
```diff
@@ -374,6 +374,11 @@ where
 
         self.write_register(Register::RegFifoAddrPtr, 0x00u8).await?;
 
+        // Clear stale IRQ flags before entering Rx. RegIrqFlags is W1C with no
+        // auto-clear on mode change (SX1276 DS §4.1.2.4); covers the listen()
+        // path, which does not call set_irq_params.
```
lthiery:
```suggestion
        // Interrupt flags stay latched until the host clears them by writing a 1
        // (SX1276 DS §4.1.2.4) — entering Rx does not reset them. Clear here so a
        // flag left over from an earlier operation can't read as a result of this
        // one; this also covers listen(), which never calls set_irq_params.
```

lthiery (changes requested):

lthiery:
Thanks for your patience @swaits 

I've been struggling to get hardware out to test many of these backed up PRs, so instead I revived a concept from a year plus ago where I wanted to match our Rust driver against the Semtech drivers in C (which became A LOT better with the SWL2001 release). I finally got this bindings crate up: https://github.com/lora-rs/smtc-modem-cores

In my PR #445, I integrate the C drivers into our unit testing (and uncovered a few bugs!) and build up some test fixtures which I think demonstrates your bug directly; see `test_restarted_rx_ignores_stale_flags`. If we land #445 you can uncomment the [#ignore] on that test.

I think I implemented fixes for @lucasgranberg 's feedback in the suggestions just now.

lthiery:
@swaits same as on #433, I used maintainer edit access to get this ready to land:

1. Rebased onto current main.
2. Applied my earlier suggestions (both clear sites go through `clear_irq_status()`, expanded the comments).
3. Un-ignored `test_restarted_rx_ignores_stale_flags`. This is an emulator test we recently landed that reproduces exactly the phantom-packet bug you're fixing: a receive that gets abandoned leaves RxDone latched, and a restarted receive reports it as a fresh packet. It fails without your fix and passes with it, so the fix is now proven in CI without hardware, which was the sticking point earlier.

lthiery (approved):

swaits:
Thanks!!!

PR #433, opened 2026-05-11
phy: add runtime LoRa sync-word setter

swaits:
Add `LoRa::set_lora_sync_word(sync_word: u8)` so the LoRa sync word can be changed on a running driver. Today the only public path for setting the sync word is at construction:

- `LoRa::with_syncword(radio_kind, sync_word: u8, delay)` — accepts an arbitrary byte but is only callable while building the driver.
- `LoRa::new(radio_kind, enable_public_network: bool, delay)` — picks between the two hardcoded LoRaWAN values: `LORAWAN_PRIVATE_SYNCWORD = 0x12` and `LORAWAN_PUBLIC_SYNCWORD = 0x34` (`lora-phy/src/lib.rs:43-46`).

That leaves no way to retune the chip's sync word after init without tearing down and reconstructing the whole `LoRa<RK, DLY>` — which is awkward in `no_std` embedded contexts because the driver owns single-use SPI / GPIO pin resources. Downstream applications that need runtime sync-word switching (custom non-LoRaWAN meshes, MeshCore-style private byte `0x12`, dynamic channel-plan reconfiguration, etc.) currently have to either:

1. Pick one sync word at build time and live with it, or
2. Reach into the chip with raw SPI ops, bypassing lora-phy.

Neither is great. This PR adds the obvious missing API.

### Background

The SX126x stores the LoRa sync word in two 8-bit registers: `LoRaSyncwordMSB` at `0x0740` and `LoRaSyncwordLSB` at `0x0741` (SX1261/SX1262 datasheet Rev 2.2 §13.4.4, register map table). `init_lora` already writes them once at cold init via the existing inline sequence at `sx126x/mod.rs:264-270`.

The SX127x family stores it in a single 8-bit register, `RegSyncWord` at `0x39` (SX1276/77/78/79 datasheet §6.4 "Digital Interface", LoRa register map). `init_lora` writes it at `sx127x/mod.rs:140` via the existing `write_register(Register::RegSyncWord, sync_word)` call.

The LR1110 / LR1121 use the `SetLoRaSyncWord` command (opcode `0x022B`) with a single u8 payload (LR1121 user manual §6.6.4). `init_lora` writes it at `lr1110/mod.rs:1343-1346`.

In all three cases the underlying chip primitive is byte-addressable. The current `init_lora(sync_word: u8)` signature already reflects that.

### Implementation

```rust
pub trait RadioKind {
    async fn init_lora(&mut self, sync_word: u8) -> Result<(), RadioError>;
    /// Apply a new LoRa sync word to the chip
    async fn set_lora_sync_word(&mut self, sync_word: u8) -> Result<(), RadioError>;
    ...
}
```

with chip-specific impls and a public `LoRa::set_lora_sync_word` wrapper.

Each chip impl is the same single-register write or single opcode that `init_lora` already performs for the same byte. Duplicated rather than factored, to keep this PR strictly additive and `init_lora` untouched. The diff is **+35 lines, –0 lines**:

```
 lora-phy/src/lib.rs        | 12 ++++++++++++
 lora-phy/src/lr1110/mod.rs |  7 +++++++
 lora-phy/src/mod_traits.rs |  2 ++
 lora-phy/src/sx126x/mod.rs | 10 ++++++++++
 lora-phy/src/sx127x/mod.rs |  4 ++++
 5 files changed, 35 insertions(+), 0 deletions(-)
```

The public wrapper at the `LoRa<RK, DLY>` level mirrors the existing discipline already used by `sleep` and `prepare_modem` for any chip- touching operation:

```rust
pub async fn set_lora_sync_word(&mut self, sync_word: u8) -> Result<(), RadioError> {
    self.radio_kind.ensure_ready(self.radio_mode).await?;
    if self.radio_mode != RadioMode::Standby {
        self.radio_kind.set_standby().await?;
        self.radio_mode = RadioMode::Standby;
    }
    self.radio_kind.set_lora_sync_word(sync_word).await?;
    self.sync_word = sync_word;
    Ok(())
}
```

Three points worth flagging:

1. **`ensure_ready` first** — SX1261/2 datasheet Rev 2.2 §8.3.1 ("Host Controller Interface", Busy management): "the host _must_ check that the `BUSY` line is at a low level before sending a new command". `ensure_ready` is the existing mechanism in this crate for that wait (either a `GetStatus` pulse if waking from Sleep, or a direct `wait_on_busy` otherwise). Matches the preamble that `sleep`, `prepare_modem`, `prepare_for_tx`, etc. all use.

2. **Driven to Standby before the register write** — both `WriteRegister` (SX126x) and the equivalent `SetLoRaSyncWord` op (LR1110) are documented as safe in Standby. Some packet-engine configuration commands on SX126x are not safe outside Standby; we apply the same discipline to keep behavior consistent. SX127x has no BUSY pin (`wait_on_busy` is a no-op) so the precondition is trivially satisfied — the standby step still runs for state- tracking consistency.

3. **`self.sync_word = sync_word` cached** — the `LoRa` struct's `sync_word` field is read by `do_cold_start` → `init_lora` during `LoRa::init()`. Any later cold reset (e.g. wedge recovery via `NRESET` + `lora.init()`) re-applies whatever value is in this field. Without the cache update the driver would silently revert to the construction-time default on recovery.

### Precedent

- **In this crate** — `LR1110::set_gfsk_sync_word(&[u8])` at `lora-phy/src/lr1110/mod.rs:712` is exactly the same pattern (runtime sync-word setter, post-construction) already established for GFSK on LR1110. This PR brings the same affordance to LoRa on SX126x / SX127x / LR1110.

- **RadioLib** — `SX126x::setSyncWord(uint16_t syncWord, uint8_t controlBits)` accepts an arbitrary 16-bit value and writes it straight to `RADIOLIB_SX126X_REG_LORA_SYNC_WORD_MSB` / `_LSB`. RadioLib defines `RADIOLIB_SX126X_SYNC_WORD_PRIVATE = 0x1424` and `RADIOLIB_SX126X_SYNC_WORD_PUBLIC = 0x3444` as named constants but does not restrict the API to those values. Same shape on `SX127x::setSyncWord(uint8_t)` (single byte to `REG_SYNC_WORD`).

- **Semtech reference drivers** — the SX126x driver published as part of the Semtech "LoRa SDK" exposes the sync-word register pair via `sx126x_set_lora_sync_word(...)` (a thin `WriteRegister` wrapper). No equivalent existed in lora-phy until now.

### Testing

Verified on hardware (Heltec V4 and Heltec Mesh Node T114):

- Two-dongle bench test rotating through six sync-word values on the TX side (`0x1424`, `0x3444`, `0x5464`, `0x7484`, `0xA4B4`, `0xC4D4` — the 8-byte register-pair encodings of bytes `0x12`, `0x34`, `0x56`, `0x78`, `0xAB`, `0xCD` per the SX126x `convert_sync_word` helper) while the RX side stays pinned to one. RX receives exactly the packets whose sync word matches its own (RSSI ≈ –58 dBm, SNR ≈ 12 dB, validated against ambient noise floor).
- MeshCore interop: with the private byte `0x12` (register pair `0x1424`) applied via this API, packets are correctly received by unmodified MeshCore repeaters using RadioLib's `RADIOLIB_SX126X_SYNC_WORD_PRIVATE` constant.
- Cold-reset survival: triggered the recovery state (NRESET + `lora.init()`) on a host application that had previously called `set_lora_sync_word(0x56)`; chip resumes on `0x56`, not on the construction-time `0x12`, confirming the `self.sync_word` cache update.

### Compatibility

Pure addition. No existing public API changed. No behavioral change to `init_lora`, `with_syncword`, or any other path. Adds one trait method to `RadioKind` (the trait is `pub` but downstream impls are expected to live in this crate's chip drivers — there is no existing out-of-tree `RadioKind` impl pattern; the trait sits behind the `LoRa<RK, DLY>` façade).

### References

- Semtech SX1261/SX1262 datasheet, Rev 2.2 (Dec 2024)
  - §8.3.1 — Host controller interface, BUSY signaling protocol
  - §13.4.4 — Register map, `LoRaSyncwordMSB` (0x0740) /
    `LoRaSyncwordLSB` (0x0741)
- Semtech SX1276/77/78/79 datasheet, Rev 7 (May 2020)
  - §6.4 — Digital interface, LoRa register map, `RegSyncWord` (0x39)
- Semtech LR1121 user manual, Rev 1.1
  - §6.6.4 — `SetLoRaSyncWord` opcode (0x022B)
- Semtech AN1200.22 — "LoRa Modulation Basics" (sync word as
  single-byte protocol concept)
- RadioLib v6.x — `SX126x::setSyncWord(uint16_t, uint8_t)` and
  `SX127x::setSyncWord(uint8_t)` for prior-art API shape
- `lora-phy/src/lr1110/mod.rs:712` — existing
  `set_gfsk_sync_word(&[u8])` as in-crate precedent

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -44,11 +44,11 @@ use interface::*;
 use mod_params::*;
 use mod_traits::*;
 
-/// Sync word for public LoRaWAN networks
-const LORAWAN_PUBLIC_SYNCWORD: u8 = 0x34;
+/// Sync word for public LoRaWAN networks (0x34 in the legacy single-byte form)
+const LORAWAN_PUBLIC_SYNCWORD: u16 = 0x3444;
 
-/// Sync word for private LoRaWAN networks
-const LORAWAN_PRIVATE_SYNCWORD: u8 = 0x12;
+/// Sync word for private LoRaWAN networks (0x12 in the legacy single-byte form)
+const LORAWAN_PRIVATE_SYNCWORD: u16 = 0x1424;
 
 /// Provides the physical layer API to support LoRa chips
 pub struct LoRa<RK, DLY>
@@ -59,7 +59,7 @@ where
     radio_kind: RK,
     delay: DLY,
     radio_mode: RadioMode,
-    sync_word: u8,
+    sync_word: u16,
     cold_start: bool,
     calibrate_image: bool,
 }
@@ -69,8 +69,14 @@ where
     RK: RadioKind,
     DLY: DelayNs,
 {
-    /// Build and return a new instance of the LoRa physical layer API with a specified sync word
+    /// Build and return a new instance of the LoRa physical layer API with a specified sync word,
+    /// given in the legacy single-byte form (e.g. 0x34 for public LoRaWAN networks)
     pub async fn with_syncword(radio_kind: RK, sync_word: u8, delay: DLY) -> Result<Self, RadioError> {
+        Self::with_syncword_raw(radio_kind, sync_word_from_legacy(sync_word), delay).await
+    }
+
+    // Build with a sync word in the 16-bit form (see [`LoRa::set_lora_sync_word`])
+    async fn with_syncword_raw(radio_kind: RK, sync_word: u16, delay: DLY) -> Result<Self, RadioError> {
         let mut lora = Self {
             radio_kind,
             delay,
@@ -95,7 +101,7 @@ where
         } else {
             LORAWAN_PRIVATE_SYNCWORD
         };
-        Self::with_syncword(radio_kind, sync_word, delay).await
+        Self::with_syncword_raw(radio_kind, sync_word, delay).await
     }
 
     /// Wait for an IRQ event to occur
@@ -187,8 +193,14 @@ where
         self.radio_kind.set_standby().await
     }
 
-    /// Apply a new LoRa sync word to the chip
-    pub async fn set_lora_sync_word(&mut self, sync_word: u8) -> Result<(), RadioError> {
+    /// Apply a new LoRa sync word to the chip.
+    ///
+    /// The sync word is given in the 16-bit form the sx126x family writes to
+    /// its sync word registers; the legacy single-byte form 0xYZ used by the
+    /// older chips corresponds to 0xY4Z4 (LoRaWAN public 0x34 -> 0x3444,
+    /// private 0x12 -> 0x1424). The sx127x and lr1110 only support values of
+    /// that shape and return `RadioError::InvalidSyncWord` for others.
+    pub async fn set_lora_sync_word(&mut self, sync_word: u16) -> Result<(), RadioError> {
         self.radio_kind.ensure_ready(self.radio_mode).await?;
         if self.radio_mode != RadioMode::Standby {
             self.radio_kind.set_standby().await?;
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -187,6 +187,18 @@ where
         self.radio_kind.set_standby().await
     }
 
+    /// Apply a new LoRa sync word to the chip
+    pub async fn set_lora_sync_word(&mut self, sync_word: u8) -> Result<(), RadioError> {
+        self.radio_kind.ensure_ready(self.radio_mode).await?;
+        if self.radio_mode != RadioMode::Standby {
+            self.radio_kind.set_standby().await?;
+            self.radio_mode = RadioMode::Standby;
+        }
+        self.radio_kind.set_lora_sync_word(sync_word).await?;
+        self.sync_word = sync_word;
+        Ok(())
+    }
+
     /// Place the LoRa physical layer in low power mode, specifying cold or
     /// warm start (if chip supports it)
     pub async fn sleep(&mut self, warm_start_if_possible: bool) -> Result<(), RadioError> {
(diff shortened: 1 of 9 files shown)


darconeous:
I haven't fully reviewed the code, but... Don't you think we should have it take the 16-bit SX126x style sync word representation instead of the 8-bit representation from the older chips?

It seems like all the newer chips use the 16-bit sync-word. I would personally prefer taking the 16-bit version, since I think it would be more future-proof.

swaits:
Yeah I think that's a nicer API; definitely matches the present and future better. I'll push another version with this change soon.

lthiery (approved):

lthiery:
@swaits I hope you don't mind, I used maintainer edit access to move this along:

1. Rebased your commit onto current main.
2. Added tests that check the setter against Semtech's reference driver (SWL2001) on all three chips. We recently landed a fixture setup for this (#455-#458), so the sync word writes are now compared byte-for-byte against the reference implementation in CI, no hardware needed.
3. Made the change @darconeous suggested: the setter now takes the 16-bit sx126x-style sync word, which is now the canonical form internally. The legacy single-byte form 0xYZ maps to 0xY4Z4 (public 0x34 = 0x3444, private 0x12 = 0x1424). The sx126x writes the full 16-bit value through, so values with no single-byte equivalent are now reachable; sx127x and lr1110 map down to their single-byte registers and return a new `RadioError::InvalidSyncWord` for values outside that shape. `LoRa::new` and `with_syncword` keep their existing signatures.

Your original commit is intact at the bottom of the branch, with my changes as separate commits on top. Once CI is green I think this is ready.

swaits:
Thanks for driving this home. I got insanely busy with $WORK. 

PR #434, opened 2026-05-20
phy: sx126x: sync host state around CAD

swaits:
## Problem

After a CAD_ONLY operation the SX126x auto-returns to STDBY_RC, but `LoRa::cad()` left `radio_mode` reading `ChannelActivityDetection` — the next `prepare_for_tx`/`prepare_for_rx` then started from an unknown host-side state (cf. RadioLib issue #1085).

## Fix

- `cad()`: explicit `SetStandby` + `radio_mode` sync on the CAD success path, matching what the error path already does.
- `do_cad()`: `SetStandby` at entry, and a defensive IRQ-flag clear before `SetCadParams`/`SetCAD` (SX1261/2 DS §8.5 — IRQ status bits are latched).

No public API changes. Pairs with #428 (pre-command BUSY check); best merged after it, though it does not depend on it to build.

## Test

clippy (`-p lora-phy --features defmt-03 -- -D warnings`) and fmt clean.

### DIFF
--- a/lora-phy/src/lib.rs
+++ b/lora-phy/src/lib.rs
@@ -460,7 +460,13 @@ where
                 .process_irq_event(self.radio_mode, Some(&mut cad_activity_detected), true)
                 .await
             {
-                Ok(Some(IrqState::Done)) => Ok(cad_activity_detected),
+                Ok(Some(IrqState::Done)) => {
+                    // CAD_ONLY exit returns the chip to STDBY_RC on its own; sync
+                    // radio_mode so the next operation starts from a known state.
+                    self.radio_kind.set_standby().await?;
+                    self.radio_mode = RadioMode::Standby;
+                    Ok(cad_activity_detected)
+                }
                 Err(err) => {
                     self.radio_kind.ensure_ready(self.radio_mode).await?;
                     self.radio_kind.set_standby().await?;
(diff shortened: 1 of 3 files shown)


lthiery:
@swaits same treatment as #432/#433, maintainer edit to get this ready to land:

1. Rebased onto current main.
2. Kept the core fix: the CAD_ONLY exit returns the chip to STDBY_RC on its own, so syncing `radio_mode` after `cad()` completes is right, and the sx126x emulator test now pins it (after `cad()` the chip is in standby and a second `cad()` without a fresh `prepare_for_cad` is rejected).
3. Dropped the two chip-level additions in `do_cad` (the entry `SetStandby` and `ClearIrqStatus`). We now test our drivers byte-for-byte against Semtech's reference driver (SWL2001) in CI, and the reference goes straight to SetCadParams/SetCAD at CAD entry, clearing IRQ flags in its handler instead; the comparison test fails with the extra commands in place. IRQ clearing on the completion path is already covered by `process_irq_event`.

With the trim the full suite is green, including the CAD byte-stream comparisons on all three chips. Once CI confirms I think this is good to go.

lthiery (approved):

swaits:
Thanks!!!

PR #440, opened 2026-07-07
device: bind RX windows to the uplink at TX time instead of recalling channel state

lthiery:
## Motivation

The region layer stored `last_tx_channel` as a side effect of TX channel selection and reconstructed RX1 parameters from it (plus the `Mac`'s *current* `data_rate`) when the RX windows opened, seconds later. Any state change in between — an interleaved certification/multicast uplink, a MAC command processed from a class C RXC frame, a channel plan mutation — could desynchronize the RX windows from the transmission they belong to. The `Frame` enum also had to be threaded through the entire RX path just so the region could guess what kind of TX it was recalling.

## Bug fixed

The recall mis-derived the RX1 datarate for fixed-plan join frames: the join TX DR is forced to DR0/DR4 by channel, but RX1 was computed from the configured data rate. A US915/AU915 join on the 500 kHz bank (one in nine attempts of the standard join walk) transmitted at DR4 while the device opened RX1 at DR10 (derived from the default DR0) instead of DR13 — a dead RX1 window; only RX2 could save the attempt. Regression test included (`test_join_rx1_datarate_follows_forced_join_dr`).

## Design

- Region TX selection (`select_tx_channel`, formerly `get_tx_dr_and_frequency`) returns a `TxChannel` carrying the DR **actually used** and the RX1 frequency paired with the selected channel.
- `Mac::join_otaa`/`send` derive an `RxWindows` (RX1+RX2 `RfConfig`) from it at TX time and hand it to the caller by value alongside the `TxConfig`.
- `async_device` keeps it as a local across the TX→RX1→RX2 sequence, so the recall's lifetime is exactly the transaction's lifetime, enforced by the borrow checker. The `nb_device` state machine carries it through its states.
- `last_tx_channel`, `get_rx_frequency(frame, window)`, and the comment-justified `unwrap` on the recalled channel in the dynamic plan are gone. Class C's RXC config remains a live query of current MAC state (via a shared `rx2_rf_config` helper), which is correct for a standing window.

## Behavior change to note (class C edge case)

RX2 parameters (RxParamSetup frequency/DR overrides) are now snapshotted at TX time. In class A this is observationally identical: a successful RX1 means RX2 never opens, so nothing can legitimately change RX2 params between TX and RX2. In class C, a MAC command processed from an RXC frame *mid-cycle* used to affect that same cycle's RX2; now it takes effect from the next uplink. Arguably more correct — the server can't assume a device applied a command it hasn't yet acknowledged — but it is a change.

## Breaking changes

None to the public API: everything touched is `pub(crate)` (`Mac`, `RegionHandler`, `TxChannel`/`RxWindows`, `nb_device::state`). `PhyRxTx`, `RfConfig`/`RxConfig`/`TxConfig`, `Device`, and the region-configuration API are unchanged.

## Testing

- New regression test for the fat-bank join RX1 datarate.
- Full suite passes including `certification` + `multicast` features (55 tests); fmt/clippy clean vs base.

### DIFF
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -310,7 +310,7 @@ where
     pub async fn join(&mut self, join_mode: &JoinMode) -> Result<JoinResponse, Error<R::PhyError>> {
         match join_mode {
             JoinMode::OTAA { deveui, appeui, appkey } => {
-                let (tx_config, _) = self.mac.join_otaa::<G, N>(
+                let (tx_config, rx_windows, _) = self.mac.join_otaa::<G, N>(
                     &mut self.rng,
                     NetworkCredentials::new(*appeui, *deveui, *appkey),
                     &mut self.radio_buffer,
@@ -325,7 +325,7 @@ where
 
                 // Receive join response within RX window
                 self.timer.reset();
-                Ok(self.rx_downlink(&Frame::Join, ms).await?.into())
+                Ok(self.rx_downlink(&Frame::Join, ms, &rx_windows).await?.into())
             }
             JoinMode::ABP { nwkskey, appskey, devaddr } => {
                 self.mac.join_abp(*nwkskey, *appskey, *devaddr);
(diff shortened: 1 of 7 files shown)


plaes (approved):
Excellent!

Issue #435, opened 2026-06-01
BaseBandModulationparams from Modulation- and Packet-Params


jakobolason:
I was wondering if it was possible to implement `From` for `BaseBandModulationParams` in `lora-modulation` for the `ModulationParams` and `PacketParams` found in `lora-phy`, or if there is a reason they are not implemented?

plaes:
Noone who needed these hasn't bothered sending patches. See also #358.

jakobolason:
I'll try and look into it in, in a weeks time, might send a patch

Issue #358, opened 2025-03-05
Dummy/Mock RadioKind?


chudsaviet:
I tried to setup unit tests for my project, and it turns out that in current lora-rs implementation its nearly impossible.

Even if I wrap LoRa struct in a trait, which partially works,  it's impossible to create ModulationParams in external crate.

So what if I contribute a DummyRadio code, which would not require any hardware to be used? Would this project accept it?

plaes:
Yes, I think this would be helpful.

PR #441, opened 2026-07-08
impl From for ease of access

jakobolason:
So this is mostly what I meant by #435  (I enjoyed vacation a bit more than anticipated, so it took a month for me to spend 3 minutes on this)

I needed an internal method to get the Baseband modulation from the already allocated SF, BW and CR, but without spending the extra place keeping them in a separate place, so this would have done nicely for me. What I ended up doing in my project was creating the BB mod params in the start, and keeping it in RAM, not great ...

I am curious about #358, which I would be happy to try and tackle, should i place it in a separate dir like the sx126x and sx127x, or is there an already good place for it?

### DIFF
--- a/lora-phy/src/mod_params.rs
+++ b/lora-phy/src/mod_params.rs
@@ -1,3 +1,4 @@
+use lora_modulation::BaseBandModulationParams;
 pub use lora_modulation::{Bandwidth, CodingRate, SpreadingFactor};
 
 /// Errors types reported during LoRa physical layer processing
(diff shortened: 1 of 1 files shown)


plaes:
Please reword the commit message to something like: `phy: Add BaseBandModulationParams -> ModulationParams` (or something better..) :)

And while at that, maybe move code just after the `struct ModulationParams` definition.

plaes (approved):

plaes:
Thanks!

> I am curious about https://github.com/lora-rs/lora-rs/issues/358, which I would be happy to try and tackle, should i place it in a separate dir like the sx126x and sx127x, or is there an already good place for it?

I guess adding it under lora-phy makes most sense. Have you also seen https://github.com/lora-rs/virtual-lorawan-device ?

PR #445, opened 2026-07-24
Test radio drivers against Semtech's reference implementation (SWL2001)

lthiery:
Tests lora-phy's radio drivers against Semtech's reference C implementation ([SWL2001](https://github.com/Lora-net/SWL2001) v4.9.0) by comparing the SPI traffic both produce for the same operations. Every lora-phy driver with a SWL2001 counterpart is covered: **sx126x**, **lr1110**, and **sx127x**. Supersedes #233 (same branch history, restructured description).

The bindings live in their own repository per earlier review feedback: [lora-rs/smtc-modem-cores](https://github.com/lora-rs/smtc-modem-cores), now **published on crates.io** (v0.1.0) and pulled in as a test-only dev-dependency — no git pin, no submodule; the sys crate packages only the compiled SWL2001 sources (255 KiB compressed). This PR is purely lora-phy test integration plus the driver bugs the comparison caught — no runtime dependencies change.

## Why

Driver encoding bugs are invisible to unit tests written from the same (mis)reading of the datasheet that produced the code. Comparing against Semtech's own driver catches them mechanically — and it did: **15 real bugs**, several of them silent on-air corruption.

### lr1110 (8, commit "fix lr1110 command encodings against the SWL2001 reference")

Most of these were SX126x conventions ported into a chip that doesn't share them:

1. **Sleep config used the SX126x bit layout** — warm start is bit 0 on lr11xx, not bit 2. Every sleep was a cold sleep, silently dropping retention (and every LoRaWAN session state kept in it).
2. **Symbol timeout used the SX126x mantissa/exponent encoding** — the lr11xx takes a raw count for values ≤255, so 8 symbols became 32 (~4× the intended RX window).
3. **WriteBuffer8 sent a stray offset byte** — the command has none; the 0x00 would have gone over the air as the first payload byte.
4. **ReadBuffer8 sent 0 in its length field** — reads returned nothing.
5. **GFSK modulation params applied SX126x chip-format conversions** — the lr11xx takes raw bps/Hz; GFSK rates were garbage.
6. **GFSK packet status read the payload length from an RSSI byte** (layout is rssi_sync, rssi_avg, rx_len, status).
7. **AutoTxRx was missing the intermediary-mode and timeout fields** — short command on the wire.
8. **The high-ACP workaround was opt-in only** — the reference applies it on every SetTx/SetRx/SetCad; affected firmware (LR1110 0x0303–0x0307, LR1120 0x0101) wasn't covered on normal paths. Now applied in do_tx/do_rx/do_cad.

### sx127x (7, commits "fix sx127x frequency, TX power and RSSI decode…", "fix the sx1276 500 kHz sensitivity errata frequency gate", "apply errata 2.3…")

1. **PLL step math truncated the low 8 bits** — fractional-MHz channels (868.1, 903.9 MHz) sat up to ~15 kHz off carrier. The existing unit test only checked whole/half-MHz values, exactly where the shortcut happened to work.
2. **PA_BOOST above 17 dBm used the PaDac-off power formula** — 18–20 dBm requests overflowed the 4-bit OutputPower field; a 20 dBm request programmed roughly 7 dBm.
3. **RFO at or below 0 dBm cast a negative OutputPower to u8** — corrupted RegPaConfig including the PA_BOOST select bit.
4. **RegDetectionOptimize RMW dropped bit 7 (AutomaticIFOn)** on every set_modulation_params.
5. **Packet RSSI skipped the 16/15 linearization in the negative-SNR branch** (reference and LoRaMac-node apply it in both).
6. **The errata 2.1 (500 kHz sensitivity) frequency gate compared kHz literals against Hz** — dead code on every real channel; the optimization values were never written.
7. **Errata 2.3 (spurious LoRa reception) was never applied** — the reference programs the manual-IF workaround on every SetRx; ours now applies it with the modulation config, including re-arming AutomaticIFOn when returning to 500 kHz.

### sx126x (1 bug — encodings otherwise verified byte-identical)

The suite largely validated the existing driver, including the §15.1 TxModulation, §15.2 TX clamp, and §15.4 inverted-IQ workarounds, and the full RX/CAD path.

**One caveat worth knowing about the TX power comparison.** SWL2001 splits responsibility differently than lora-phy does: its driver only encodes SetPaConfig/SetTxParams, while the actual power table (which pa_duty_cycle/hp_max/power triple to use per dBm — datasheet Table 13-21) lives in its BSP layer, which this PR doesn't compile. So the reference is fed our driver's hardcoded triples, and the comparison verifies the command encoding while pinning our table as test data — it does not independently validate the table values against Semtech's choices. Two table-level differences are documented in test comments rather than "fixed":

- The 14 dBm-and-below high-power row turned into a driver fix during review of this PR. It's the one place ST's STM32WL table differs from datasheet Table 13-21 (checked against Rev 2.2, Dec 2024): same 0x02/0x02 PA config, but ST commands the target dBm directly where the datasheet interpolates SetTxParams from its +22 setpoint. Both tables are correct for their own hardware — the STM32WL is an SX126x die integrated into ST's package, characterized by ST — but the driver applied ST's value to every high-power variant, so **discrete SX1262 parts got the wrong row** (bug 15 of this PR, making the sx126x count 1 rather than 0). Fixed with `Sx126xVariant::pa_table() -> &'static PaTable` — the PA table is const data the variant supplies (flash, no RAM): `Sx1262`/`Sx1261` default to the datasheet tables, `Stm32wl` returns ST's, and boards with their own PA characterization can plug in their own table. Both shipped tables are pinned by comparison tests. A small fix fell out of the table refactor too: a 21 dBm request now interpolates to SetTxParams 21 within the +22 row instead of over-commanding 22.
- Image calibration: SWL2001's generic `cal_img_in_mhz` (ceil of band-edge/4) lands one step short of datasheet Table 9-2 for the 902–928 MHz band (0xE8 vs 0xE9). Our driver follows the table; the tests compare against the reference's fixed-byte `cal_img` entry point with the table values, and the lr11xx equivalent has the same off-by-one.

Also documented upstream: `sx127x_set_ocp_value` (sx127x side) always programs 45 mA — it masks the trim argument with the inverted-form `OCP_TRIM_MASK`, zeroing whatever is passed.

## How the comparison works

Each chip needed its own fixture shape, matching how its drivers talk:

- **sx126x** (command-based): both drivers record transactions into one op log; equality is wire-canonical (trailing-NOP-trimmed writes + total bytes clocked) because the two drivers split read transactions differently while clocking identical wire bytes.
- **lr1110** (command-based, two-transaction reads): both drivers split reads identically, so transactions compare exactly. Canned responses key off the preceding command write; command-less direct reads consume a FIFO.
- **sx127x** (register-based): the drivers legitimately factor traffic differently (burst RMW + shadow registers vs single-register RMW over SPI), so both execute against a small register file and equality is the chip-visible outcome — final register state plus the FIFO byte stream. The fixture models the OpMode LongRangeMode latch the reference silently relies on.

Read-path tests also assert decoded-value parity (RSSI/SNR/lengths), not just bytes.

## Emulator

`lora-phy/src/sx126x/test/emulator.rs` is a behavioral sx126x model that the **real** `Sx126x` driver and `LoRa` state machine run against unmodified — tests hold a `Chip` handle to inject received packets and inspect transmitted ones. End-to-end tests cover TX, RX, sleep/wake, CAD (quiet and busy channels), and single-mode RX timeout with recovery. Relates to #442: this tests actual driver code where a trait-level mock tests none, at the cost of being chip-specific; the two approaches complement each other.

## Bindings design (smtc-modem-cores)

The crate owns the `#[no_mangle]` HAL symbols the C drivers link against and dispatches through a monomorphized vtable stored at the head of a `repr(C)` context, so `Context::new(spi)` is the whole user API and multiple SPI types coexist in one binary. Each chip family's HAL needed its own shape: sx126x (write+read), lr11xx (write, two-transaction read, direct read, plus no-op GPIO-level symbols), sx127x (register write/read hung off `sx127x_t.hal_context`, host timers stubbed).

Beyond testing lora-phy, the bindings could anchor development of an sx128x driver (SWL2001 ships one; lora-phy has none).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,10 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+      # Needed to build the smtc-modem-cores bindings used by lora-phy tests
+      - name: Setup apt packages
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends cmake llvm-dev clang libclang-dev
+
       - name: Setup Cache
         uses: Swatinem/rust-cache@v2
         with:
@@ -87,6 +91,10 @@ jobs:
         uses: styfle/cancel-workflow-action@0.12.1
         with:
           access_token: ${{ github.token }}
+
+      # Needed to build the smtc-modem-cores bindings used by lora-phy tests
+      - name: Setup apt packages
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends cmake llvm-dev clang libclang-dev
 
       - name: Setup Cache
         uses: Swatinem/rust-cache@v2
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "70c6e87" }
+smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "42d8559" }
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "1949152" }
+smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "f2b684b" }
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -36,4 +36,6 @@ lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
 smtc-modem-cores = "0.2.1"
+# Derive `OpCode::from_repr` for the sx126x behavioral emulator (test-only)
+strum = { version = "0.28", default-features = false, features = ["derive"] }
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = { git = "https://github.com/lora-rs/smtc-modem-cores", rev = "b31ab4a" }
+smtc-modem-cores = "0.1.0"
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
-smtc-modem-cores = "0.1.0"
+smtc-modem-cores = "0.2.1"
 tokio = { version = "1", features = ["rt", "macros"] }
--- a/lora-phy/Cargo.toml
+++ b/lora-phy/Cargo.toml
@@ -35,5 +35,5 @@ lorawan-radio = ["dep:lorawan-device"]
 lorawan-device = { path = "../lorawan-device" }
 # Bindings to Semtech's reference driver (SWL2001), used to test our
 # implementation against the vendor's SPI byte stream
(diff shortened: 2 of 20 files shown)


lthiery:
@plaes @lucasgranberg If either of you guys want to review this, give me a heads up, otherwise I might land it sooner than later. Sorry for the amount of code but its like 80% test fixtures.

plaes:
How hard would it be to split this up into:
* base stuff (Common test framework)
* sx127x 
* sx126x
* lr111x

Kezii:
if you decide to split, I tested all LR1110 changes of this branch to be working on real hardware

lthiery:
> How hard would it be to split this up into:
> 
> * base stuff (Common test framework)
> * sx127x
> * sx126x
> * lr111x

The commit history makes it pretty easy to untangle. There's very little actual common lines of code for testing, to be clear; it's just the methodology that's the same. Anyway, I'll open 4 fresh PR's unrolling the idea. 1 base PR, 3 PRs against base

Kezii:
can't the fixes be made in another PR against main? the test fixture is uncorrelated to the actual driver fixes

lthiery:
> can't the fixes be made in another PR against main? the test fixture is uncorrelated to the actual driver fixes

The individual driver fixture kind of motivates the fix, so it kind of makes sense to bundle them. But let me think about that over breakfast

lthiery:
Ok, this is how it looks broken up:

- #455: the chip-independent scaffolding, against main (smtc-modem-cores dev-dependency, shared fixture boilerplate, CI packages)
- #456: sx126x comparison tests, emulator and the PA power table rework
- #457: sx127x comparison tests, sx1276 errata fixes and the DIO0 emulator
- #458: lr1110 comparison tests, command encoding fixes and vendor PA power tables

The three module PRs are based on #455 and are independent of each other; GitHub will retarget them to main once the base merges. Each module's driver fixes stay interleaved with the tests that forced them, and every commit builds and passes tests on its own. Merging all four gives a tree identical to this PR's head.

Leaving this one open for reference until those land.

lthiery:
@Kezii I think I like the way this is now chunked. You should be able to continue any testing or development against PR #458 (ie: branch `lthiery/lr1110-swl2001`)

Issue #449, opened 2026-07-29
LR1110 unable to join a LoRaWAN network due to bugs


Kezii:
I am trying to use this library with LR1110 for LoRaWAN

Several bugs were ecountered along the way, including #444, #448 and others

Most of the stuff should be fixed in #445, but for now in my testing the radio is not seeing downlink messages yet

I think this issue should remain open and track progress until someone manages to get an LR1110 to join LoRaWAN

lthiery:
Working on a 445 rebase. I'll ping you here when it's up

lthiery:
Ok take a look! It's rebased

Kezii:
Thank you @lthiery 

For now, I confirm the previous findings

It's able to transmit a correct joinrequest (so, unlike the current main, the 0x00 offset etc is fixed)
The gateway transmits a join response, but it's not seen by the radio

```
[20.458160] test_lora.rs INFO  trying to join
[20.500823]     mod.rs DEBUG set_irq_params: dio1_mask = 0xffffffff
[20.503631]     mod.rs DEBUG sf = 12, bw = 4, cr = 1
[20.505767]     mod.rs DEBUG channel = 868300000
[20.508117]     mod.rs DEBUG set_irq_params: dio1_mask = 0x00000404
[22.049957]     mod.rs DEBUG get_irq_flags: stat1=05, stat2=53, flags=0x00000004
[22.050048]     mod.rs DEBUG process_irq: irq_flags = 0x00000004 in radio mode Transmit
[22.050628]     mod.rs DEBUG Starting RX1 in 4950 ms.
[27.000640]     mod.rs DEBUG Configuring RX1 window with config RxConfig { rf: RfConfig { frequency: 868300000, bb: BaseBandModulationParams { sf: _12, bw: _125KHz, cr: _4_5, ldro: true, t_sym_us: 32768 }, max_payload_len: 59 }, mode: Single { ms: 50 } }.
[27.041748]     mod.rs DEBUG set_irq_params: dio1_mask = 0xffffffff
[27.044555]     mod.rs DEBUG sf = 12, bw = 4, cr = 1
[27.045806]     mod.rs DEBUG channel = 868300000
[27.046386]     mod.rs DEBUG set_irq_params: dio1_mask = 0xffffffff
[27.049865]     mod.rs DEBUG get_irq_flags: stat1=04, stat2=59, flags=0x00000000
[...]

PR #450, opened 2026-07-29
busy guard for lr1110, fix race condition

Kezii:
this change was proposed by my random LLM while seeing this issue https://github.com/lora-rs/lora-rs/issues/449#issuecomment-5122266027

please check that the change is harmless

I'm testing with a NRF52832

we are still dependent on #445 for lorawan to work correctly on the LR1110

I tested that device.join( ) returns JoinResponse::JoinSuccess, no traffic after that

### DIFF
--- a/lora-phy/src/lr1110_interface.rs
+++ b/lora-phy/src/lr1110_interface.rs
@@ -31,6 +31,9 @@ where
 
     // Write a buffer to the radio.
     pub async fn write(&mut self, write_buffer: &[u8], is_sleep_command: bool) -> Result<(), RadioError> {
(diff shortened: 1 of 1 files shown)


lthiery:
> I tested that device.join( ) returns JoinResponse::JoinSuccess, no traffic after that

Were you expecting more traffic? Assuming you're using Class A and not C, you will need to send Uplinks to get anything down. If you send a confirmed uplink that's a simple way of showing the basics work.

Kezii:
yeah no I was not expecting more traffic, it was just to state what I tested to be working

I just tested uplink and downlink too

lthiery (approved):

PR #457, opened 2026-07-30
phy: test the sx127x driver against SWL2001

lthiery:
Part of the #445 split, based on #455 (GitHub will retarget to main once that merges).

The sx127x suite: register-level comparison tests that run our driver and Semtech's reference implementation (SWL2001) against the same register-file fixture and assert the writes match. Since the sx127x is register-based rather than command-based, the fixture keeps an ordered register write log so sequencing bugs show up too.

Driver fixes the comparisons forced:
- frequency, TX power and RSSI decode corrected against SWL2001
- the sx1276 500 kHz sensitivity errata frequency gate was wrong
- errata 2.3 (spurious LoRa reception) applied on the sx1276

Also included: a behavioral emulator with edge-triggered DIO0 so the full TX/RX flow runs under the async driver, and a test pinning every Register discriminant to SWL2001's register values.

Each commit builds and passes tests on its own.

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -1,0 +1,28 @@
+name: HIL bench
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [hil-tests]
+  pull_request:
+    branches: [hil-tests]
+
+concurrency:
+  group: hil-bench
+  cancel-in-progress: false
+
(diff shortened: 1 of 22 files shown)


plaes:
Btw, we also support sx1272 :)

PR #460, opened 2026-07-30
lorawan-device: reject out-of-range max data rate in NewChannelReq

lthiery:
## Problem

In dynamic-channel regions (EU868/EU433/AS923/IN865), `handle_new_channel` validates the DataRateRange from a received NewChannelReq by iterating `min..=max` and indexing `datarates()`, which has 15 entries (DR0..=DR14). `DataRateRange::can_build_from` only rejects `max < min`, so a range with max data rate 15 (for example wire byte 0xFF, meaning min 15 / max 15) indexes `datarates()[15]` and panics. A single crafted downlink can crash the device.

## Fix

Reject ranges whose max data rate is at or above `NUM_DATARATES` before iterating, and NAK the command.

## Tests

Regression test uses wire byte 0xFF (range 15..=15), which reaches the bad index directly. Note that 0xF0 (range 0..=15) does not trigger the panic on EU868 because the iteration short-circuits on DR6 being unsupported, so the test uses the direct trigger. Verified the test panics with an index-out-of-bounds on the code before the fix, plus an acceptance case for a supported range.

### DIFF
--- a/lorawan-device/src/region/dynamic_channel_plans/mod.rs
+++ b/lorawan-device/src/region/dynamic_channel_plans/mod.rs
@@ -311,8 +311,12 @@ impl<R: DynamicChannelRegion> RegionHandler for DynamicChannelPlan<R> {
 
         // Check if DataRateRange is valid and supported
         if let Some(r) = dr {
-            let dr_supported = (r.min_data_rate()..=r.max_data_rate())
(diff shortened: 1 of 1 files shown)


plaes (approved):

plaes (dismissed):
Thanks. Introduced in d797bb254cf45b36ecc5c28f577bc5793c4ceb3d

PR #461, opened 2026-07-30
Apply DLSettings from JoinAccept and check MIC before applying state

lthiery:
Two fixes in JoinAccept handling:

1. The DLSettings byte (RX1DROffset and RX2DataRate) was parsed but never applied (the TODO in `otaa.rs`), so a server that configures either value in the accept was ignored and subsequent downlinks could be missed. Values are now validated against the region the same way RXParamSetupReq handling does it; invalid values keep the current settings.

2. The MIC was checked after region state was already updated: a JoinAccept that failed MIC validation still processed the CFList / channel mask and set RX1 delay. The MIC check now happens before any state changes. One of the new tests demonstrates the old behavior: a bad-MIC accept carrying RxDelay=3 moved rx1_delay from 1000 to 9000 ms even though the join was rejected.

Three new tests: valid DLSettings applied, invalid values ignored (defaults kept), bad-MIC accept changes nothing.

### DIFF
--- a/lorawan-device/src/async_device/test/mod.rs
+++ b/lorawan-device/src/async_device/test/mod.rs
@@ -100,6 +100,76 @@ async fn test_no_join_accept() {
 }
 
 #[tokio::test]
+async fn test_join_accept_dl_settings_applied() {
+    let (radio, timer, mut async_device) = setup();
+    // Run the device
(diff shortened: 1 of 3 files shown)


plaes (approved):
Excellent!

PR #462, opened 2026-08-03
Add hardware-in-the-loop test crate (tests-hil-fw)

lthiery:
Adds an on-target test crate that runs lorawan-device OTAA scenarios on real hardware: a NUCLEO-WL55JC talking over the air to a US915 gateway (RAK2287), with a network-side fixture acting as the join server. The paired scenarios run in the \"HIL bench\" workflow on a self-hosted runner (sandboxed, one job per throwaway container; outside PRs need a first-run approval).

What the tests cover so far:

- Full OTAA join plus a decrypted first uplink, both sides asserting (device exit code and fixture events)
- Negative joins: MIC-tampered JoinAccept and no-answer, device must never join and no heartbeat may decrypt
- JoinAccept DLSettings adoption (hardware check of #461): nonstandard RX1DROffset/RxDelay and RX2 datarate, verified by receiving a downlink placed where the accept said to listen. The RX1 variant fails against pre-#461 code, so it genuinely discriminates the fix.

Notes for review:

- The crate is workspace-excluded: no_std, its own lockfile, and embedded-test pinned to 0.6.x (embassy-stm32 0.2 is version-locked to embassy-executor 0.7).
- The cargo runner assumes the bench (probe-rs remote on the gateway Pi); on any other machine the tests simply can't run, they are CI/bench-only.
- Repeatability on the bench: 10/10 consecutive joins on attempt 0, and every push to this branch has run the full 5-scenario suite green.

Draft while the remaining scenarios land (RX2 fallback, FCnt/JoinNonce persistence across resets). Opening now for early eyes on the crate layout and the runner shape.


Kezii:
does it make sense to crowdsource the HIL testing? What are the requirements and implications?

I could host nrf52832 + LR1110 and a SX1262 SoC tests with EU frequencies

lthiery:
> does it make sense to crowdsource the HIL testing? What are the requirements and implications?
> 
> 
> 
> I could host nrf52832 + LR1110 and a SX1262 SoC tests with EU frequencies

I can document the rig soon if you want to run one. It's essentially a Raspberry Pi + SX1302 at its base and I plug dev boards into it via USB.

I bought [one of these RF switches on eBay](https://www.minicircuits.com/pdfs/ZN4PD1-50+.pdf) so I can have everything directly wired with attenuators but relatively isolated but that's optional. Antennas and attenuators is probably fine.

Long term, I can try to think of a solution so we could "crowdsource" the CI integration. I like the idea bc if you want better coverage for a board, bring it!

PR #465, opened 2026-08-07
lorawan-encoding: borrowed-view rewrite of parser and creator

lthiery:
Two commits that rewrite lorawan-encoding's parse and create surface in place: `parser.rs`, `creator.rs` and `maccommands.rs` are modified at their existing paths, not moved to a new module.

Commit 1 introduces the new API side by side with the old one and asserts byte-for-byte equivalence on the crate's shared test vectors: parsing, frame building (MICs included), join session-key derivation, and MAC command framing. Commit 2 deletes the old surface, moves the new one into its place, and migrates lorawan-device (MAC core, regions, the whole async_device test suite). Net -3000 lines. To audit the equivalence claims, review commit 1 in isolation.

## What goes away

The old design threaded a `T: AsRef<[u8]>` storage generic through every parser type and hung shared behavior off traits. All of it is replaced with plain modern Rust:

| removed | replaced by |
|---|---|
| `T: AsRef<[u8]>` generic on every view type; a fresh monomorphization of the whole parse/decrypt pipeline per storage type | borrowed `&'a [u8]` views; one monomorphization total |
| `AsPhyPayloadBytes` trait (plus an `impl AsRef<[u8]> for dyn AsPhyPayloadBytes`) | an inherent `as_bytes()` on each view |
| `MICAble` / `MHDRAble` blanket impls that index into the buffer on every call | validation once at the parse boundary (a private `Layout` records offsets), so accessors are infallible by construction |
| `DataHeader` trait with default methods re-deriving fhdr length and re-slicing per call | plain inherent methods over the validated layout |
| `fixed_len_struct!` wire fields generic over storage (`DevAddr<T>`, `EUI64<T>`, ...) with cross-storage `PartialEq<DevAddr<T>> for DevAddr<V>` and `to_owned()` conversions | owned `[u8; N]` newtypes: `Copy`, `Eq`, `Hash`, derived `PartialEq`, no conversions needed |
| setter-chain creators where some setters return `&mut Self` and others `Result<&mut Self>`, breaking the chain | struct literals with `..Default::default()`; the validation the type system can't express happens once in `build_into` |
| `MacCommandIterator<'a, T>` yielding bare commands and silently stopping at malformed input | `Iterator<Item = Result<Cmd, ParseError>>`, fused after the first error; `ParseError` implements `Display` and `core::error::Error` |

The types also get stricter where the spec allows it: `DevEui` and `JoinEui` are distinct types so swapping them in a JoinRequest is a compile error, `NonZeroU8` FPort makes application data on port 0 unrepresentable, the `Payload` enum makes MAC-commands-plus-data unrepresentable, and FPort 0 with non-empty FOpts is rejected per spec (the old creator allowed it). Byte order is decided in one place: wire fields store transmission order and convert explicitly, where the old `From<u32> for DevAddr` used big-endian array order and round-tripped through the LSB-first wire with flipped endianness.

Decryption keeps the encrypted/decrypted typestate but drops the storage generic: the caller hands the `&mut [u8]` to `decrypt_in_place` / `check_mic_and_decrypt_in_place` explicitly.

## Bugs that die with the old framing

The legacy iterator panicked on a lone McGroupStatusAns CID byte (reachable from received frame contents), fabricated an empty EchoIncPayloadAns from a lone 0x08, and silently dropped truncated tails. It also compared minimum payload lengths against the data including the CID byte; `TxFramesCtrlReqPayload::min_len` had been calibrated to that off-by-one, so correct counting initially rejected the spec's one-octet FrameType frame. The device certification tests caught it and `min_len` is now 1.

Deliberate behavior changes, each covered by a side-by-side test in commit 1: `f_port()` returns `Some` when an FPort byte is present with an empty FRMPayload (the old parser conflated this with no-FPort), all parse entry points reject a non-R1 major version, and malformed MAC command tails yield `Err` instead of ending the stream silently.

## Validation

Commit 1 kept both APIs alive and asserted byte-for-byte equivalence on the shared vectors. With the old oracle gone, those assertions are self-contained: creators still reproduce the wire vectors exactly, decryption is pinned as an involution over the same vectors, and derived session keys are pinned to values captured while both APIs coexisted. 178 encoding tests + 73 device tests green, all feature combinations checked (serde, defmt-03, certification, multicast, class-c, no-default-features).

The crate's criterion bench (ported to the new API in place) shows the validate-once design is also faster on the host, x86 numbers from the same machine, old API at the merge-base vs this branch: header parsing + field access 10.5 ns to 6.5 ns (-38%), decrypt 54.2 ns to 50.4 ns (-7%), MIC validation unchanged at ~138 ns (CMAC dominates). Zero heap allocations per iteration in both, as before.

Flash cost went down, not up: `size-compare/` builds the same parse+decrypt+join workload for thumbv7em-none-eabi, and the new API is smaller in every configuration (652 vs 874 text bytes at one storage type with fat LTO, 1442 vs 1872 at three; the old API paid ~250 bytes per additional storage type, the new one ~20-40 per extra entry point).

## lorawan-device migration

Mostly mechanical. Three API consequences: `Session` / `SessionKeys` / `JoinMode::ABP` use the owned `DevAddr`, the region trait's `process_join_accept` takes `Option<&CfList>` instead of being generic over the join-accept buffer, and multicast uses the owned `McAddr` plus a raw-key `decrypt_in_place_with` for its session keys.


lorawan-encoding/src/certification.rs
```diff
@@ -236,8 +268,11 @@ impl<'a> TxFramesCtrlReqPayload<'a> {
         Ok(TxFramesCtrlReqPayload(data))
     }
 
+    /// Minimum length of the payload not including CID (one FrameType octet).
+    /// The retired iterator compared this against the data including the CID
```
plaes, line 272: Do we really need this historical data?

lthiery: Ah no - not at all. I admittedly was mostly paying attention to the code only when I pushed this thru and didn't do a good job with comments. Thanks for pointing out the slop! I just landed another PR to try to clean it up from the commit. It was a lot of LLM churn but I think it was a worthwhile overhaul.

Please keep any and all feedback coming! I appreciate it

PR #474, opened 2026-08-09
increase rx window by a bit, fix missing downlinks on lr1110 with faster data rates

### LINKED ISSUE #473: LR1110 unable to downlink a packet with packet parameters different than DR::_0
as title, it seems to fail, while SX1262 works

Kezii:
fixes #473 

I don't know why this problem only appears on lr1110, as far as I can see loramac-node does something similar (albeit more complex)

### DIFF
--- a/lora-phy/src/lorawan_radio.rs
+++ b/lora-phy/src/lorawan_radio.rs
@@ -169,7 +169,13 @@ impl RxMode {
                 // Since both sx126x and sx127x have a preamble-based timeout, we translate
                 // the additional millisecond delay into symbols and add it to the amount of preamble symbols.
                 const PREAMBLE_SYMBOLS: u16 = 13; // 12.25
-                let num_symbols = PREAMBLE_SYMBOLS + bb.delay_in_symbols(ms);
+
+                // add a fixed margin to the timeout to account for the late arrival of the packet in fast modulation modes
+                const MIN_LATE_MARGIN_MS: u32 = 250;
+                let timeout_ms = ms
+                    .saturating_add(bb.symbols_to_ms(PREAMBLE_SYMBOLS as u32))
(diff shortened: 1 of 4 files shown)


lora-phy/src/lorawan_radio.rs
```diff
@@ -169,7 +169,13 @@ impl RxMode {
                 // Since both sx126x and sx127x have a preamble-based timeout, we translate
                 // the additional millisecond delay into symbols and add it to the amount of preamble symbols.
                 const PREAMBLE_SYMBOLS: u16 = 13; // 12.25
-                let num_symbols = PREAMBLE_SYMBOLS + bb.delay_in_symbols(ms);
+
+                // add a fixed margin to the timeout to account for the late arrival of the packet in fast modulation modes
```
plaes, line 173: nit: Sentence should start with capital letter.

lora-phy/src/lorawan_radio.rs
```diff
@@ -169,7 +169,13 @@ impl RxMode {
...
+                const MIN_LATE_MARGIN_MS: u32 = 250;
```
lthiery, line 174: This feels like an aggressive addition. It looks like loramac-node defaults a similar value to 10ms. Let's maybe start with that but make it configurable?

lthiery (changes requested):
I opened up an alternative implementation that tries to go a bit further in adopting the timing strategy established in Loramac Node. It adopts their symbol-based window scaling and offsets and exposes something similar to your `MIN_LATE_MARGIN_MS` as `set_system_max_rx_error`

I'd be curious if you could test out #479 and see if it works for you @Kezii (btw, I should hopefully have the LR111x kit hooked up my HIL this week - I realized yesterday I had ordered the wrong thing)

Kezii:
a more careful fix in #479

PR #475, opened 2026-08-09
Add Adaptive Data Rate

Kezii:
Adds initial Adaptive Data Rate support.

ADR is enabled by default and can be controlled through Device::set_adr / get_adr. Uplinks set ADR and ADRACKReq as appropriate, accepted downlinks reset the acknowledgement counter, and the data rate backs off after ADR_ACK_LIMIT + ADR_ACK_DELAY missed downlinks.

This is an incremental implementation and does not yet support NbTrans, so it is not the complete LoRaWAN 1.0.4 ADR behavior.

### DIFF
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -301,6 +301,28 @@ where
         self.mac.configuration.data_rate = datarate;
     }
 
+    /// Whether Adaptive Data Rate (ADR) is enabled.
+    ///
+    /// When enabled, uplinks set the FCtrl ADR bit so the network may adjust
+    /// data rate and TX power via LinkADRReq. ADRACKReq / DR backoff also run
+    /// when connectivity is lost.
+    pub fn get_adr(&self) -> bool {
+        self.mac.configuration.adr_enabled
+    }
+
+    /// Enable or disable Adaptive Data Rate (ADR).
+    ///
+    /// ADR is enabled by default. Disable it for mobile devices or when the
+    /// application manages data rate itself.
+    pub fn set_adr(&mut self, enabled: bool) {
+        self.mac.configuration.adr_enabled = enabled;
+        if !enabled {
+            if let Some(session) = self.mac.get_session_mut() {
+                session.adr_ack_cnt = 0;
+            }
+        }
+    }
+
     /// Join the LoRaWAN network asynchronously. The returned future completes when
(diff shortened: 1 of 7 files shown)


lorawan-device/src/async_device/mod.rs
```diff
@@ -301,6 +301,28 @@ where
...
+        self.mac.configuration.adr_enabled
+    }
+
+    /// Enable or disable Adaptive Data Rate (ADR).
+    ///
+    /// ADR is enabled by default. Disable it for mobile devices or when the
+    /// application manages data rate itself.
+    pub fn set_adr(&mut self, enabled: bool) {
+        self.mac.configuration.adr_enabled = enabled;
+        if !enabled {
```
plaes: I wonder if we could reduce nesting.. This would[ require Rust 1.88+](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains) though
```rust
    if !enabled && let Some(session) = self.mac.get_session_mut() {
        session.adr_ack_cnt = 0;
    }
```

plaes (commented):
Looks good, though I would need to run it against the certification tool. Might take few days though.

lorawan-device/src/async_device/mod.rs
lthiery: I'm all for bumping the rust edition

lthiery:
I'll try to spend a moment early this week reviewing too.

FWIW, my HIL test rig is coming along. I have the ST integrated SX126x dev board hooked up atm with proper attenuators. I have their SX127x and LR1110 ones on hand too but waiting on some more RF connectors and board rework before it's all set.

Would be nice to try to encode the certification tests into the suite eventually...

plaes:
Apparently LCTT tests require `nbtrans` support in order to test this.

lorawan-device/src/async_device/mod.rs
lthiery: https://github.com/lora-rs/lora-rs/pull/478

lthiery (dismissed):
I bumped the rust edition and version upstream so once you rebase, this should be good for landing as a step towards improving ADR.

lthiery (approved):
Rebased onto current main, review feedback addressed, ADR threshold/backoff coverage added, and CI is green. Good to land as the incremental ADR step described in the PR.

Issue #312, opened 2024-12-28
Sx127x IRQ Confusion


martin2250:
Hi,

I'm trying to make an RFM92 Module (Sx1276) work on my custom board, controlled by an stm32l053r8 with embassy. However, the application seems to lock up every time it expects to receive a packet and no packet arrives.

## What I Did

I used the [STM32L0 example](https://github.com/lora-rs/lora-rs/blob/main/examples/stm32l0/src/bin/lora_lorawan.rs), only changing the device keys, region, the pin mapping and adding a loop to actually send data at the end of the main function:
```rust
    // only for example, I tried all DIOs. used in the `GenericSx127xInterfaceVariant`
    let dio0_pin = Input::new(p.PB1.degrade(), Pull::Up);
    let dio0_irq = ExtiInput::new(dio0_pin, p.EXTI1.degrade());
    // at the end of main, after the join
    loop {
        device.send(&[0xde, 0xad, 0xbe, 0xef], 1, false).await.unwrap();
        Timer::after(Duration::from_secs(30)).await;
    }
```
For the IRQ pin I wasn't sure which DIO to use (my board exposes all to the STM32), so I tried all of them. For some, the transmit didn't succeed. If that worked, the application would get completely stuck:
- when the join accept message wasn't received
- when nothing was received during the first RX window (even unconfirmed messages)

## Debugging

I believe that the issue is that with the current way IRQs are handled by the SX127x implementation, the timeout interrupt is not received. In `set_irq_params`, the mask is manipulated for both the DIO0 and DIO3 pins. We're only able to configure a single pin, however.

I got everything to work perfectly by spoofing the irq pin like this:
```rust
struct Irq;

impl embedded_hal::digital::ErrorType for Irq {
    type Error = core::convert::Infallible;
}

impl embedded_hal_async::digital::Wait for Irq {
    async fn wait_for_high(&mut self) -> Result<(), Self::Error> {
        Timer::after_millis(100).await;
        Ok(())
    }

    async fn wait_for_low(&mut self) -> Result<(), Self::Error> {
        defmt::todo!()
    }

    async fn wait_for_rising_edge(&mut self) -> Result<(), Self::Error> {
        defmt::todo!()
    }

    async fn wait_for_falling_edge(&mut self) -> Result<(), Self::Error> {
        defmt::todo!()
    }

    async fn wait_for_any_edge(&mut self) -> Result<(), Self::Error> {
        defmt::todo!()
    }
}
```
[...]

PR #476, opened 2026-08-09
phy: sx127x: optionally watch DIO1 for RxTimeout

lthiery:
## Problem

`GenericSx127xInterfaceVariant` watches a single IRQ pin (DIO0), which carries RxDone/TxDone/CadDone. The sx127x routes **RxTimeout only to DIO1**, and `lorawan_device::async_device::Device` needs it to close an empty receive window. With DIO0 alone, an RX window that hears nothing never wakes the driver, so an OTAA join (or any receive) that gets no downlink hangs forever instead of moving on to the next window/attempt.

The struct doc already conceded this ("not compatible with `async_device::Device` due to its reliance on `RxTimeout` ... on secondary IRQ pin"), so single-DIO boards had no working generic LoRaWAN variant.

## Fix

Add `new_with_secondary_irq(reset, irq, irq_secondary, rf_switch_rx, rf_switch_tx)`. When a DIO1 pin is supplied, `await_irq` wakes on whichever of DIO0/DIO1 fires (via `embassy_futures::select`, DIO0 polled first so a coincident RxDone is reported ahead of RxTimeout); the driver then reads `RegIrqFlags` to tell the events apart, exactly as before. `new()` is unchanged (secondary = `None`) and keeps its single-pin behavior for LoRa P2P and CAD.

This pulls in `embassy-futures` as a dependency, already used across the `examples/` in this repo.

## Testing

Hardware-validated on a B-L072Z-LRWAN1 (STM32L072CZ + SX1276, single-DIO wiring, US915) against a real gateway and network-server fixture. Full OTAA suite 5/5: join, DLSettings RX1/RX2 downlink reception, and, the load-bearing cases, two negative scenarios where the server never sends a JoinAccept. Those only terminate because RxTimeout on DIO1 closes each empty window; on the old single-DIO variant they hang. Reverting to `new()` reproduces the hang.

## Compatibility

Additive, non-breaking: existing `new()` callers are untouched.

### DIFF
--- a/lora-phy/CHANGELOG.md
+++ b/lora-phy/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- sx127x: Add `GenericSx127xInterfaceVariant::new_with_secondary_irq` to watch DIO1 (RxTimeout), fixing LoRaWAN RX-window hangs on single-IRQ boards
 - Bump MSRV to 1.75
 - Add documentation for crate features
 - Various documentation fixes (link rot, update docstrings to reflect reality)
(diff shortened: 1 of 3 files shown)


plaes (approved):
Thanks! Also fixes #312

PR #485, opened 2026-09-08
Implement NbTrans support (Number of Redundant Uplink Transmissions)

plaes:
Tested against LCTT for EU868 and US915 using virtual-lorawan-device.

Now `TP_A_EU868_ED_MAC_104_BV_012_A.Redundancy` and `TP_A_US915_ED_MAC_104_BV_012_A.Redundancy` tests pass.

### DIFF
--- a/lorawan-device/src/async_device/mod.rs
+++ b/lorawan-device/src/async_device/mod.rs
@@ -371,7 +371,9 @@ where
     /// receives no downlink in RX1/RX2 is retransmitted automatically with the same FCntUp,
     /// up to NbTrans times (1..=15). Each retransmission is a full transmission with its
     /// own channel selection and RX1/RX2 windows, and a downlink in either window
-    /// stops the retransmission.
+    /// stops the retransmission. This also applies to uplinks the stack itself
+    /// transmits (such as certification answer frames on FPort 224), which open
+    /// their own RX windows after transmission.
     ///
     /// In Class C mode, it is possible to get one or more downlinks and `Reponse::DownlinkReceived`
     /// maybe not even be indicated. It is recommended to call `take_downlink` after `send` until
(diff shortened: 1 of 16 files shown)


lthiery:
Reviewed this with the help of a bot on the HIL infrastructure and I think it came up with a few valid points:

1. Retries should not impact ADRACKCnt

TS001 1.0.4 section 4.3.1.1: "Each time the uplink frame counter is incremented (for each new uplink frame, because repeated transmissions do not increment the frame counter), the end-device SHALL increment an ADRACKCnt counter."

I think the fix is `if retransmit {..}` around `self.retransmissions += 1` inside of `lorawan-device/src/mac/session.rs::rx2_complete`.

2. ADR backoff never resets NbTrans to 1

LinkADRReq can raise this to 15, but nothing brings it back down when the network goes quiet. TS001 1.0.4 section 4.3.1.1 (Table 9, the row at ADRACKCnt 160 and above) has the last backoff step reset NbTrans to 1 once the device is at the default data rate and ADR_ACK_DELAY more uplinks went unanswered. Without it a device stuck at DR0 keeps sending every frame NbTrans times for nothing.

Suggested change: in the backoff step of rx2_complete, when next_lower_datarate returns None, set configuration.nb_trans = 1. The channel re-enable from the same sentence needs a region API that does not exist yet, fine as a follow-up.

3. Confirmed uplinks are retransmitted without RETRANSMIT_TIMEOUT

Section 4.3.1.3: after a confirmed uplink with no ACK the device SHALL wait RETRANSMIT_TIMEOUT (RP002: 1-3 s random) after RECEIVE_DELAY2 before the next uplink, repetition or new.

4 Class C: certification answer sent in the RX1 gap plus NbTrans > 1 leaves inconsistent state

`async_device/mod.rs::rx_downlink`: The gap handler calls rx2_complete() once and drops the result. With nb_trans > 1 that returns Retransmit and leaves retransmissions at 2, retransmit_buffer holding the answer, and pending_cert_rx never taken. When the original uplink's RX2 then times out, the send loop retransmits the answer instead of finishing. Suggest not arming retransmission for answers sent in the gap (reset session.retransmissions there) or looping rx2_complete until it stops returning Retransmit.

5. Certification answer RX windows open slightly late (async)

The stashed TX duration is fine, but timer.reset() happens back in send() after rx_listen returns and window_complete() has run, not right after the answer's TX done. Milliseconds on real radios, so probably acceptable; mentioning in case the LCTT timing checks are tight.

plaes:
Ok, simpler things hopefully fixed:

1. [x] Retries should not impact ADRACKCnt - 0d6094a9e707ac2d415a71f4cbfec0044c6423ca
2. [x] **(*)** ADR backoff never resets NbTrans to 1 - 0659e13fcda37c54e6f79a9cd71a53f17441079b - suggested channel handling in #488 
3. [ ] Confirmed uplinks are retransmitted without RETRANSMIT_TIMEOUT
4. [x] Class C: certification answer sent in the RX1 gap plus NbTrans > 1 leaves inconsistent state  - 898407a73c3ecd0f23f5fac961881fd448dcaf71
5. [ ] Certification answer RX windows open slightly late (async)

And comments:
- 3rd one might take longer as well, as it spans almost all layers: regions + mac + async + nb
- 5th one is hopefully optional? :)

lthiery (approved):
Nice work! And yea we can carve out the remaining items as follow on.

PR #486, opened 2026-09-19
lorawan-device: non-volatile store and LoRaWAN 1.0.4 join rules

### LINKED ISSUE #335: LoRaWAN 1.0.4 support (DevNonce storage).
Our current implementation for `DevNonce` handling does not conform to LoRaWAN 1.0.4:
> `DevNonce` is a counter starting at 0 when the end-device is initially powered up and
incremented with every Join-Request. A `DevNonce` value SHALL never be reused for a given
JoinEUI value. If the end-device can be power-cycled, then `DevNonce` SHALL be
persistent (e.g., stored in a non-volatile memory). Resetting `DevNonce` without changing
JoinEUI will cause the Join Server to discard the Join-Requests of the end-device. For each
end-device, the Join Server keeps track of the last `DevNonce` value used by the end-device
and ignores Join-Requests if `DevNonce` is not incremented.

Current code generates new nonce's randomly:
https://github.com/lora-rs/lora-rs/blob/5fc1e96aa91fe3e1d17f7f4fa02423bc188e024c/lorawan-device/src/mac/otaa.rs#L38-L43

### LINKED ISSUE #425: Possible security improvement
Hi, apologies for how this is framed, I've been a bit around the houses here thinking about the DevNonce Prng use in `lorawan-device`, because successively generating a `u16` randomly is going to lead to collisions quite rapidly.

# Summary

I did some experiments and found that both in `lorawan-device, and actually, using any PRNG no matter how strong, the DevNonce generated will collide after on average 320 JoinRequests.

I'm open to writing a pull request with an interface to function as a monotonic counter as a non-default alternative to the DevNonce Prng and wanted to check if it's welcome. This would be in line with Lora Alliance's [Technical Recommendations for Preventing State Synchronization Issues around LoRaWAN® 1.0.x Join Procedure](https://resources.lora-alliance.org/technical-recommendations/technical-recommendations-for-preventing-state-synchronization-issues-around-lorawan-1-0-x-join-procedure) which recognise this is a problem with the 1.0.x specification more generally.

# Details

I wrote the following experiment code because I was curious if the `fastrand` wyhash implementation used by lorawan-device would possibly collide faster than it ought to. Note that `collide_joinrequests` follows the pattern used for each join request iterating on an initial nonce value, `collide_newseeds` follows the patten of initialising a `new_session_with_seed` for a set of random u64 seeds. `collide_sysrng` was included to demonstrate what the average value is with reasonably high quality system random.

```rust
#![feature(random)] // requires nightly build
use std::collections::HashSet;
use std::random::random;

const ITERATIONS: usize = 100000;

// Generate u16 randomly using fastrand until
// hit collision
fn collide_joinrequests() -> usize {
    let mut seeds: Vec<u64> = vec![];
    for _i in 0..ITERATIONS {
        // Using /dev/urandom from linux as a source of unlikely 
        // to collide seeds.
        seeds.push(random(..));
    }

    let mut results: Vec<usize> = vec![];

    for seed in seeds {
        let mut frand = fastrand::Rng::with_seed(seed);
        let mut set: HashSet<u16> = HashSet::new();
        loop {
            let new_r = frand.u32(..) as u16;
            if set.contains(&new_r) {
                // Collision hit
                results.push(set.len());
                break;
            } else {
                set.insert(new_r);
            }
        }
    }
    
    results.iter().sum::<usize>()/ITERATIONS

}

// Experiment to see how often newly instantiated fastrand with /dev/urandom
// quality seed will result in a DevNonce collision
fn collide_newseeds() -> usize {
    let mut results: Vec<usize> = vec![];
    for _i in 0..ITERATIONS {
        let mut set: HashSet<u16> = HashSet::new();
        loop {
            // New rng with new random seed
            let mut frand = fastrand::Rng::with_seed(random(..));
            let r = frand.u32(..) as u16;
            if set.contains(&r) {
                // Collision
                results.push(set.len());
                break;
            }
            set.insert(r);
        }
    }

    results.iter().sum::<usize>()/ITERATIONS
}

fn collide_sysrng() -> usize{
    let mut results = vec![];
    for _i in 0..ITERATIONS {
        let mut set: HashSet<u16> = HashSet::new();
        loop {
            // Standard /dev/urandom without fastrand
            let new_r: u16 = random(..);
            if set.contains(&new_r) {
                // Collision hit
                results.push(set.len());
                break;
            } else {
                set.insert(new_r);
            }
        }        
    }
    results.iter().sum::<usize>()/ITERATIONS
}

fn main() {
    let r1 = collide_joinrequests();
    let r2 = collide_newseeds();
    let r3 = collide_sysrng();
    println!("collide_joinrequests: Average collisions over {} runs: {}", ITERATIONS, r1);
    println!("collide_newseeds: Average collisions over {} runs: {}", ITERATIONS, r2);
    println!("collide_sysrng: Average collisions over {} runs: {}", ITERATIONS, r3);
}
```

# Results

Both the patterns used by DevNonce in lorawan-device and random chance result in a collision after ~320 invocations.

```shell
$ cargo +nightly run --release
warning: unused manifest key: package.channel
   Compiling fastrand-collider v0.1.0 (fastrand-collider)
    Finished `release` profile [optimized] target(s) in 0.21s
     Running `target/release/fastrand-collider`
collide_joinrequests: Average collisions over 100000 runs: 320
collide_newseeds: Average collisions over 100000 runs: 319
collide_sysrng: Average collisions over 100000 runs: 320
```

This isn't necessarily extremely bad (how often does a fixed installation issue a Join Request?) but it is much worse than the lifetime number of join requests or new session initialisations available if the DevNonce is set to a number, and then incremented (a lifetime of 65535 different join requests).

My proposal here is that I submit a TechNotePrng implementation of some kind where `new_session_with_seed` can be called, and the seed value can be persisted by the device using lorawan-device, and the "Prng" implementation simply increments on calling for the next value.

lthiery:
Adds an optional non-volatile store to `async_device::Device`. With a store the device follows the LoRaWAN 1.0.4 join rules and resumes its session across power loss. Without one (the default, `NoNvm`) nothing changes: existing constructors keep the 1.0.2 behavior and the persistence code compiles out.

## What a store gives you

- DevNonce is a monotonic counter, committed to the store before each join request goes on the air. A counter that reaches `u16::MAX` refuses further joins with `Error::DevNonceExhausted` rather than reusing a value.
- A JoinAccept carrying the same JoinNonce as the last accepted one is rejected as a replay. 1.0.4 only guarantees a non-repeating JoinNonce, so equality is the strongest check that never rejects a compliant server.
- Session keys, frame counters and negotiated MAC state (data rate, RX1/RX2 settings, TX power, ADR flag) survive a reboot. `Device::restore` loads them.
- FCntUp uses a checkpoint ahead of the live counter: the session blob is rewritten before transmitting a counter that reaches the checkpoint, and boot resumes from the checkpoint. A power cycle can skip up to `margin` counter values (default 32) but never reuses one. `set_checkpoint_margin` trades flash writes against skipped values, `checkpoint()` flushes at the exact counter before a planned power-down.
- ABP sessions persist too. Activating again with the same keys and address after a restore keeps the counters; new keys start a fresh session.

## Store contract

`NonVolatileStore` (async) and `NonVolatileStoreSync` (blocking, with a blanket impl) expose two regions, `Identity` and `Session`, as opaque blobs with a CRC-protected header. A torn or stale blob is treated as absent. Surviving power loss mid-write is the store's job; identity writes in particular must be atomic, since a lost identity restarts the DevNonce at 0 and a join server then rejects the device until the counter passes its last value.

## Not in this PR

- `nb_device` is not wired (it stays at 1.0.2 behavior).
- Channel-plan state and multicast sessions are not persisted. Adding them is a schema version bump.

The first commit is a small standalone fix: region data rate lookups indexed the table without a bounds check, which panics on a DR past the end.

Tested with the host suite across all feature sets and on a STM32WL55 with an internal-flash store (cross-reset join and session resume, on a separate branch).

Closes #335, closes #425.


lorawan-device/src/nvm.rs
```diff
@@ -0,0 +1,440 @@
...
+    SchemaVersion,
+    /// Blob belongs to the other region (slots swapped by the store).
+    RegionMismatch,
+    /// Length field disagrees with the buffer or the expected payload size.
+    BadLength,
+    /// Payload failed its integrity check (torn or corrupted write).
+    BadCrc,
+}
+
+const MAGIC: [u8; 4] = *b"LWNV";
```
plaes, line 186: `LWRS`? :) Aka `LoRa[WAN]-rs`

plaes (approved):
Looks good at first glance.
I need to port my certification-device + virtual-lorawan-device branches over to this one and see how well it actually works :)

PR #487, opened 2026-09-19
lora-phy: report CRC and header errors from rx() instead of handing the buffer up

kamilrybacki:
## What

On sx126x, sx127x and lr1110 a received packet whose payload CRC failed raised `RxDone` together with `CrcErr`, and `get_irq_state` only logged the second flag: `complete_rx()`/`rx()` returned the length and status of a packet whose bytes were noise, and the caller had no way to tell. A header CRC failure was logged the same way and otherwise ignored, so the caller kept waiting.

This PR treats both as outcomes of the receive:

- `RadioError::CrcError` when `CrcErr` is set (sx126x, lr1110) or `RxDone` arrives with the CRC flag (sx127x);
- `RadioError::HeaderError` when `HeaderErr` is set (sx126x, lr1110; sx127x has no such IRQ).

In continuous receive the radio stays armed, as it already does for other errors, so a caller simply loops on `rx()`; in single/timed receive the existing error path applies.

## Tests

The sx126x behavioural emulator can now inject a packet with a fate (intact, payload CRC failed, header CRC failed) and raises the flags DS.SX1261-2 describes for each (`RxDone|CrcErr`, or `HeaderErr` alone). Two end-to-end tests pin the driver's answer; the intact path is unchanged and every existing test passes (`cargo test -p lora-phy --lib`: 111 passed). `cargo fmt --check` and `cargo clippy -p lora-phy --tests -- -Dclippy::all` are clean.

## Compatibility

Two new variants on `RadioError`, which is not `#[non_exhaustive]`: an exhaustive `match` downstream will need two arms. Callers that only propagate the error see a new error where they previously received corrupted bytes as a packet. Changelog entry added under Unreleased.

## Context

Found while running this driver unmodified against a behavioural SX1262 model in a LoRa P2P protocol test harness: corrupted frames reached the application layer and were rejected only by its AEAD seal. RadioLib and the vendor driver both surface the CRC result to the caller.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BEWLR4RX6DV42WZEgZCXnY


lthiery (approved):
Thanks, this is a nice improvement and I like how you extended the mock. It breaks some error handling in the lorawan layer, but I'll follow up on that.

PR #488, opened 2026-09-21
ADR backoff - re-enable default channels in final step

plaes:
According to ADR backoff procedure, once minimum datarate has been reached and `ADRACKCnt` >= 160, device should re-enable all default channels.

We initially missed this while reviewing #475, but fortunately it came up with nbtrans implementation #485.

cc @Kezii 

I also have a feeling that `ADRACKCnt` steps are not followed, but I'll look into this once nbrans support lands so I can test it against LCTT.

image


lthiery (approved):
Looks good, approving.

Checked it against TS001 1.0.4 section 4.3.1.1 and Table 9: dynamic plans get the region defaults back and leave NewChannelReq channels alone, fixed plans go back to all channels. That matches the text (lines 604-607). Tests pass with `--all-features` here too.

On the ADRACKCnt steps you mention: Table 9 has TX power going back to default at 96 and the data rate only stepping down at 128, while we step the data rate at 96 and never touch TX power. So I think your feeling is right. Happy to review that as a follow-up whenever you get to it, no need to hold this one.

